PR #1, opened 2021-10-04
docs: Added incomplete draft of ADR 001

inventhouse:


docs/decisions/0001-coodinator-design.rst
```diff
@@ -0,0 +1,81 @@
...
+============
+
+
+
+Highly modular
+==============
+
+The world of online commerce is rapidly changing, and with so many moving pieces it is important to Open edX operators to be able to use the services and tools that are the best fit for them, in the ways that they need to. As such we have decided to use a system of decoupled events to allow maximum flexibility in implementation. The exact details of the event mechanisms will be addressed in subsequent ADRs.
+
+The system will have a fixed set of default messages handling most current use cases (ex: Payment Received, Refund Requested) and a built in set of message handlers that can be enabled via configuration which interact with other Open edX systems (LMS notified upon Payment Received). Additional Django Apps can be installed in the usual fashion which to offer additional handlers.
```
inventhouse, line 38: This makes it sound like these messages and handlers would be part of the core, that is not my expectation.

docs/decisions/0001-coodinator-design.rst
```diff
@@ -0,0 +1,81 @@
...
+
+In addition to handlers, the system will provide default URL endpoints to receive incoming API calls / callbacks from external services, and allow extensibility via the usual Django app mechanisms to add new API endpoints and callback handlers which can, in turn, fire off custom messages.
```
inventhouse, line 40: This makes it sound like the core will vend URLs / APIs, again that is not my expectation.

docs/decisions/0001-coodinator-design.rst
bmedx: I'm not sure what you mean by "core" in this context. In our discussion about whether to have multiple repositories it sounded like these pieces of functionality would be part of this repository, and so included in the distribution?

PR #3, opened 2021-10-29
docs: ADR-2 on using asynchronous processing and Celery

bmedx:
**Description:** Adding a new ADR that will codify our decision to use an asynchronous processing model for Commerce Coordinator based upon Celery.

**JIRA:** REV-2404

### DIFF
--- a/docs/decisions/0002-asynchronous-processing.rst
+++ b/docs/decisions/0002-asynchronous-processing.rst
@@ -1,0 +1,28 @@
+############################
+0002 Asynchronous Processing
+############################
+
+Status
+******
+
+**Draft**
+
+Context
+*******
+
+The Commerce Coordinator is most significantly a broker for network calls to other services. This pattern can introduce latency and frailty into the entire service when a downstream service is experiencing performance or availability issues. EdX has experience mitigating these kinds of issues by using a message queue and separating the processing of the networking requests from the processing of APIs and other low-latency work using Celery queues and tasks.
+
+Decisions
+*********
+
+- This application will primarily utilize an asynchronous message queue to handle communication with other services by default.
+- This application will utilize Celery as the message broker.
+- Celery tasks will be run in the context of Celery worker processes, not in the API-providing Commerce Coordinator service.
+- Code for API endpoints, Celery tasks, and task handlers should be co-located in one place and released together.
+
+Consequences
+************
+
+Unless there is a very compelling reason not to, all message handling that calls to external services from the Commerce Coordinator will be done in Celery tasks. Those tasks will run on a different set of servers / in different processes than the ones which power the Commerce Coordinator API, allowing the API to be resilient to downstream service issues.
+
+Developing, debugging, and testing Celery tasks can be more challenging than writing synchronous code. Efforts will be made early to simplify these things. Where possible, shared code will handle things like logging, tracing, and testing tasks. When not possible, thorough documents will be written explaining how to do the different things necessary to develop and administer these tasks.
--- a/docs/decisions/0002-asynchronous-processing.rst
+++ b/docs/decisions/0002-asynchronous-processing.rst
@@ -18,6 +18,7 @@ Decisions
 - This application will primarily utilize an asynchronous message queue to handle communication with other services by default.
 - This application will utilize Celery as the message broker.
 - Celery tasks will be run in the context of Celery worker processes, not in the API-providing Commerce Coordinator service.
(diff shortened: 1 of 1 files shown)


docs/decisions/0002-asynchronous-processing.rst
```diff
@@ -0,0 +1,28 @@
...
+The Commerce Coordinator is most significantly a broker for network calls to other services. This pattern can introduce latency and frailty into the entire service when a downstream service is experiencing performance or availability issues. EdX has experience mitigating these kinds of issues by using a message queue and separating the processing of the networking requests from the processing of APIs and other low-latency work using Celery queues and tasks.
+
+Decisions
+*********
+
+- This application will primarily utilize an asynchronous message queue to handle communication with other services by default.
+- This application will utilize Celery as the message broker.
+- Celery tasks will be run in the context of Celery worker processes, not in the API-providing Commerce Coordinator service.
+- Code for API endpoints, Celery tasks, and task handlers should be co-located in one place and released together.
+
```
dianekaplan, line 22: ^ These last two bullets are a little confusing to me. Could you say a little more about what'll be suitable to go into an API endpoint vs a celery task? Or do you mean we'll have API endpoints that kick off celery tasks rather than doing the work themselves?

docs/decisions/0002-asynchronous-processing.rst
```diff
@@ -0,0 +1,28 @@
...
+Consequences
+************
+
+Unless there is a very compelling reason not to, all message handling that calls to external services from the Commerce Coordinator will be done in Celery tasks. Those tasks will run on a different set of servers / in different processes than the ones which power the Commerce Coordinator API, allowing the API to be resilient to downstream service issues.
```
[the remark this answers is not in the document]
bmedx: It should be, this is how (I believe) all of our non-ecomm-worker Celery services run.

docs/decisions/0002-asynchronous-processing.rst
bmedx: Thanks for the feedback, I'll add some more info here. I meant the latter, but I can definitely clarify it.

dianekaplan: that's helpful- thanks!

inventhouse (approved):
It may not _have_ to be in this ADR, but we will need to figure out the details about celery tasks sending signals at some point

docs/decisions/0002-asynchronous-processing.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Developing, debugging, and testing Celery tasks can be more challenging than writing synchronous code. Efforts will be made early to simplify these things. Where possible, shared code will handle things like logging, tracing, and testing tasks. When not possible, thorough documents will be written explaining how to do the different things necessary to develop and administer these tasks.
+
+Regarding code location, an extension may live in a Github repository external to Commerce Coordinator, but should have the code for all related API endpoints and Celery tasks that it manages inside that repository. This allows releasing all related code together at roughly the same time, even though the services are running in different processes.
```
julianajlk, line 31: What kind of things should be done in an extension? Is it basically anything that is not related to API endpoints and Celery tasks? Or it's just that these things should always be in the Commerce Coordinator repo but other things could outside of it, or not

bmedx: The expectation is that extensions can be located in any repo, or any place where pip can install it from. They are basically Django apps, so they can do anything a Django app can such as:

- Create an API endpoint URL
- Add an admin UI
- Create models in the database (discouraged, but may be necessary for things like configuration)
- Create new Signals
- Receive existing Signals
- Create Celery tasks

The Commerce Coordinator core is really just plumbing and helpers to gather the extensions together and make writing them easier / more consistent. The next ADR should clarify all of this, I hope.

@julianajlk  Given that I'm writing a much more detailed ARD about the CC internals, do you think we need to put more information here, or is this good for now?

bmedx:
> It may not _have_ to be in this ADR, but we will need to figure out the details about celery tasks sending signals at some point

@inventhouse I dug into this a little bit, and it seems that as long as we're importing things the same way in the Celery workers as we do in the main app this should work for us: https://stackoverflow.com/questions/3920174/django-signals-in-celery

docs/decisions/0002-asynchronous-processing.rst
julianajlk: This is good for now, makes sense to go into detail on the other ADR. Thanks @bmedx

PR #4, opened 2021-11-05
docs: ADR-3 - Internal Communication

bmedx:
**Description:** Adding a new ADR to codify use of Django Signals as a default internal communication mechanism and Django Apps as the recommended way to extend Commerce Coordinator functionality. This explicitly leaves out example code and API documentation in favor of adding those into official docs as they are built.

**JIRA:** REV-2405

### DIFF
--- a/docs/decisions/0003-internal-communication.rst
+++ b/docs/decisions/0003-internal-communication.rst
@@ -1,0 +1,77 @@
+###########################
+0003 Internal Communication
+###########################
+
+Status
+******
+
+**Draft**
+
+
+Context
+*******
+
+Per previous ADRs, Commerce Coordinator is intended to be a highly decoupled, modular, asynchronous platform for handling a wide variety of use cases and workflows. In order to allow for flexibility in configuration and extensibility the CC needs to have a loosely coupled internal architecture that allows for pluggable components to work with the same events.
+
+
+Decisions
+*********
+
+- This application will recommend using Django applications as extension points
+- This application will use Django signals as the primary means of communicating between extensions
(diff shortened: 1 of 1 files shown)


inventhouse (approved):
Looks good to me 👍

docs/decisions/0003-internal-communication.rst
```diff
@@ -0,0 +1,77 @@
...
+
+While any Python package can be installed and called from inside the Commerce Coordinator, using Django applications as extension points enables several key pieces of functionality:
+
+- Extensions can add API endpoints for triggering workflows and callbacks
+- Extensions can add Admin UI functionality
+- Extensions can store data in custom Django models (discouraged, but could be useful for configuration)
+- Extensions can create new Django signals
+- Extensions can take advantage of our custom Django signal enhancements (see below)
+- Extensions will be configurable via standard Django methodology and deployed in well understood ways
+
```
dianekaplan, line 39: Sorry for the late comment- (and I'm not even sure that others would need it in the ADR)- but for my own understanding: 
I think of most of the things in this list as normal django features. By 'Extensions' are we talking about the [django-extensions](https://github.com/django-extensions/django-extensions) package? Does that make those features easier to implement than by doing them in django?

bmedx: @dianekaplan This is really just an argument for why we want to do things the "Django way" instead of using lower level Python extension mechanisms like Stevedore extensions. It's nothing to do with django-extensions, though I can see why that's confusing.

PR #24, opened 2022-04-14
REV-2575: add orders app with endpoint to retrieve order info from ecommerce

dianekaplan:
**Description:** Add an orders app that'll be used for communication around order retrieval. The first endpoint will be to request order information for a user, but in the future we'll add another (or modify this one) to be able to call an external provider for a user's orders. 

**JIRA:** [Link to JIRA ticket](https://openedx.atlassian.net/browse/REV-2575)

**Installation instructions:** 
- follow the setup steps in the commerce-coordinator README file [here](https://github.com/edx/commerce-coordinator)
- run commerce-coordinator locally

**Testing instructions:**
Test the endpoint: with this branch, go to the new endpoint test page, specifying a user that has orders in your local environment: http://localhost:8140/orders/order_history/?username=edx&page=1&page_size=20

endpoint_output

Test via order history page:  With [this branch](https://github.com/openedx/frontend-app-ecommerce/pull/181) in progress, the order history page can go through this new endpoint to get the orders info. 
Order history via CC endpoint: 
order_history_via_CC_endpoint

Order history via ecommerce: 
order_history_via_ecommerce

### DIFF
--- a/commerce_coordinator/apps/orders/clients.py
+++ b/commerce_coordinator/apps/orders/clients.py
@@ -2,12 +2,13 @@
 API clients for services that manage orders.
 """
 import logging
-import requests
 
+import requests
 from django.conf import settings
 from edx_rest_api_client.client import OAuthAPIClient
 
 logger = logging.getLogger(__name__)
+
 
 class BaseEdxOAuthClient:
     """
@@ -38,7 +39,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
     """
     API client for calls to the edX Ecommerce service.
     """
-    api_base_url = settings.ECOMMERCE_URL + '/api/v2/'
+    api_base_url = str(settings.ECOMMERCE_URL) + '/api/v2/'
 
     def get_orders(self, username):
         """
@@ -63,4 +64,3 @@ class EcommerceApiClient(BaseEdxOAuthClient):
         except requests.exceptions.HTTPError as exc:
             logger.exception(exc)
             raise
-
--- a/commerce_coordinator/apps/orders/clients.py
+++ b/commerce_coordinator/apps/orders/clients.py
@@ -1,0 +1,66 @@
+"""
+API clients for services that manage orders.
+"""
+import logging
+import requests
+
+from django.conf import settings
+from edx_rest_api_client.client import OAuthAPIClient
+
+logger = logging.getLogger(__name__)
+
+class BaseEdxOAuthClient:
+    """
+    API client for calls to the other edX services.
+    """
+
+    def __init__(self):
+        self.client = OAuthAPIClient(
+            settings.SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT.strip('/'),
+            self.oauth2_client_id,
+            self.oauth2_client_secret,
+            timeout=(
+                settings.REQUEST_CONNECT_TIMEOUT_SECONDS,
+                settings.REQUEST_READ_TIMEOUT_SECONDS
+            )
+        )
+
+    @property
+    def oauth2_client_id(self):
+        return settings.BACKEND_SERVICE_EDX_OAUTH2_KEY
+
+    @property
+    def oauth2_client_secret(self):
+        return settings.BACKEND_SERVICE_EDX_OAUTH2_SECRET
+
+
+class EcommerceApiClient(BaseEdxOAuthClient):
+    """
+    API client for calls to the edX Ecommerce service.
+    """
+    api_base_url = settings.ECOMMERCE_URL + '/api/v2/'
+
+    def get_orders(self, username):
+        """
+        Call ecommerce API overview endpoint for data about an order.
+
+        Arguments:
+            username: restrict to orders by this username
+        Returns:
+            dict: Dictionary represention of JSON returned from API
+
+        example response:
+        {
+            # TODO: Fill in.
+        }
+        """
+        try:
+            query_params = {'username': username}
+            endpoint = self.api_base_url + 'orders/'
+            response = self.client.get(endpoint, params=query_params)
+            response.raise_for_status()
+            return response.json()
+        except requests.exceptions.HTTPError as exc:
+            logger.exception(exc)
+            raise
+
--- a/commerce_coordinator/apps/orders/clients.py
+++ b/commerce_coordinator/apps/orders/clients.py
@@ -41,7 +41,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
     """
     api_base_url = str(settings.ECOMMERCE_URL) + '/api/v2/'
 
-    def get_orders(self, username):
+    def get_orders(self, query_params):
         """
         Call ecommerce API overview endpoint for data about an order.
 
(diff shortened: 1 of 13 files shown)


commerce_coordinator/settings/base.py
```diff
@@ -84,7 +84,9 @@ def root(*x):
 CORS_ALLOW_HEADERS = corsheaders_default_headers + (
     'use-jwt-cookie',
 )
-CORS_ORIGIN_WHITELIST = []
+CORS_ORIGIN_WHITELIST = [
+    'http://localhost:1996'
+]
```
dianekaplan, line 90: TODO: Follow up and see if this is the correct pattern/approach (I don't see this elsewhere)- for now want to get the endpoint working/tested (and with an empty set, requests were getting a Network Error)

commerce_coordinator/apps/orders/tests.py
```diff
@@ -0,0 +1,46 @@
...
+        })
```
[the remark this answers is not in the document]
inventhouse: Agreed about having a relatively authentic response for the test.  I would lean toward breaking it out so it can be more readable, perhaps as a constant just above the test case; this can help as a reference when trying to understand what a response should look like.  You can automatically format it by copying that line into a python interpreter and using `pprint` like this:
```python
from pprint import pp
results = [{"billing_address":...
pp(results)
```
Some of those lines may _still_ end up being too long, but the overall "shape" of it will be easier to grok.

commerce_coordinator/apps/orders/urls.py
```diff
@@ -0,0 +1,12 @@
...
+Orders app URLS
+"""
+
+from django.urls import path
+
+from commerce_coordinator.apps.orders.views import get_user_orders__ecommerce
+
+app_name = 'orders'
+urlpatterns = [
+    path('ecommerce/', get_user_orders__ecommerce, name='orders_ecommerce'),
```
inventhouse, line 11: I would make this path something like `orders/` or `order_history/`; this API will be the generic endpoint and won't have anything to do with legacy ecommerce once we split that backend query out into its own component/app.  May also want to remove "ecommerce" from the `name` (though we might leave it in the view name, since that view currently does rely on legacy ecommerce.)  Alternatively, these changes could be part of the work to split frontend API from backend query.

commerce_coordinator/apps/orders/clients.py
```diff
@@ -0,0 +1,82 @@
...
+            (and ID verification)", "price": "149.00", "expires": "2022-11-08T22:54:30.777313Z",
+            "attribute_values": [{"name": "certificate_type", "code": "certificate_type", "value":
+            "verified"}, {"name": "course_key", "code": "course_key", "value":
+            "course-v1:edX+DemoX+Demo_Course"}, {"name": "id_verification_required", "code":
+            "id_verification_required", "value": true}], "is_available_to_buy": true, "stockrecords":
+            [{"id": 3, "product": 3, "partner": 1, "partner_sku": "8CF08E5", "price_currency": "USD",
+            "price_excl_tax": "149.00"}]}}], "number": "EDX-100004", "payment_processor":
+            "cybersource-rest", "status": "Complete", "total_excl_tax": "149.00", "user":
+            {"email": "edx@example.com", "username": "edx"}, "vouchers": []} ]}
+        }
```
inventhouse, line 73: I'd format this to be more readable or refer the reader to the fixture in `tests.py` (the former is convenient for the reader, but more likely to get stale; as-written, the outermost curly-braces should probably be square-brackets)

commerce_coordinator/apps/orders/views.py
```diff
@@ -0,0 +1,27 @@
...
+def get_user_orders__ecommerce(request):
+    """
+    To start: get orders from ecommerce for hardcode
+    """
+    username = request.GET['username']
+    page = request.GET['page']
+    page_size = request.GET['page_size']
+    params = {'username': username, "page": page, "page_size": page_size}
+
+    logger.info(f'DKTEST: get_user_orders__ecommerce called with params: {params}.')
```
inventhouse, line 22: Stray debug line?

commerce_coordinator/settings/base.py
inventhouse: I believe this should go in `settings/local.py`; CORS is different in stage and prod because they have real domains.

commerce_coordinator/settings/local.py
```diff
@@ -4,6 +4,7 @@
 
 INSTALLED_APPS += (
     'commerce_coordinator.apps.demo_lms.apps.LmsConfig',
+    # 'commerce_coordinator.apps.orders.apps.OrdersConfig',
```
inventhouse, line 7: Should this be commented out like this?

provision-commerce-coordinator.sh
```diff
@@ -1,30 +1,9 @@
...
-# Create the database
-docker exec -i commerce-coordinator.db mysql -u root -se "CREATE DATABASE commerce-coordinator;"
-
-# Run migrations
-echo -e "${GREEN}Running migrations for ${name}...${NC}"
-docker exec -t commerce-coordinator.app bash -c "cd /edx/app/${name}/ && make migrate"
-
 # Create superuser
 echo -e "${GREEN}Creating super-user for ${name}...${NC}"
-docker exec -t commerce-coordinator.app bash -c "echo 'from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser(\"edx\", \"edx@example.com\", \"edx\") if not User.objects.filter(username=\"edx\").exists() else None' | python /edx/app/${name}/manage.py shell"
```
inventhouse, line 27: We may not want to delete these things, or, at least, we may want them back when we do move commerce-coordinator into a container

inventhouse (commented):
Overall excellent work; a little feedback to consider, let me know if you have questions or want to chat about any of it.

commerce_coordinator/settings/base.py
dianekaplan: Ah of course- that makes sense! Moved and confirmed that frontend-app-ecommerce can still load the order history page via commerce-coordinator after the update

commerce_coordinator/settings/local.py
dianekaplan: Oops sorry- I wasn't sure if I'd need that, and it looks like I don't- removed!

commerce_coordinator/apps/orders/views.py
dianekaplan: Indeed! This is still helpful as I add the other test, but removed for now since it's temporary.

commerce_coordinator/apps/orders/tests.py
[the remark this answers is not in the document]
dianekaplan: Okay sure- I pulled it out to the top of tests.py now. Just to confirm, the pprint is just meant for use in the terminal/while debugging/etc? (not to be used in this file somehow?)

commerce_coordinator/apps/orders/clients.py
dianekaplan: Good call- this shouldn't be repeated! I updated the comment to refer to the sample_response in tests.py if the reader wants to see an example, good call

commerce_coordinator/apps/orders/tests.py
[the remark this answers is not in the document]
dianekaplan: Done- thank you!

commerce_coordinator/apps/orders/urls.py
dianekaplan: Agreed- thanks!

provision-commerce-coordinator.sh
inventhouse: I think leaving them commented-out in the file with a brief comment saying to put them back when we move development into docker would be best, it's easy to lose track of an external doc somewhere.

inventhouse (approved):
Recommend keeping the docker stuff in the file, otherwise looks good!

pshiu (approved):
Looks good to me, nice work Diane! Just an observation that I think right now, anyone can query the order status of any other user, so we may need to figure out authentication in a later PR.

commerce_coordinator/apps/orders/tests.py
```diff
@@ -0,0 +1,102 @@
...
+    Verify endpoint availability for order retrieval endpoint(s)
+    """
+    maxDiff = None
+
+    def setUp(self):
+        super().setUp()
+        self.mock_response = sample_response
+
+    @patch('commerce_coordinator.apps.orders.clients.EcommerceApiClient.get_orders')
+    def test_EcommerceApiClient(self, mock_response):
```
julianajlk, line 88: It can be confusing in terms of consistency throughout the whole project to use both underscore and camel case in a function name?

provision-commerce-coordinator.sh
dianekaplan: ah I like that idea- thanks Phil!

commerce_coordinator/apps/orders/tests.py
julianajlk: https://peps.python.org/pep-0008/#function-and-variable-names
^ Not sure if it applies to test functions

dianekaplan: goooooood question..... I'd imagined I should use the name of the thing I'm testing (which has a camel case name), but you're right: I'm looking at examples in ecommerce it looks like we keep camel case for the test class name (if applicable) and understores for test names- good catch! Update coming...

dianekaplan:
> Looks good to me, nice work Diane! Just an observation that I think right now, anyone can query the order status of any other user, so we may need to figure out authentication in a later PR.

Thanks Phil- you are correct! I just checked manually, and even though the ecommerce OrderViewset filter_queryset gives a PermissionDenied if a non-staff user tries to see another's orders, you're right: when I log in as a new user locally and look up orders for an admin, I do see the results! http://localhost:8140/orders/order_history/?username=edx&page=1&page_size=20

Should I add this example to the following authentication ticket, or is this a little different? 
https://openedx.atlassian.net/browse/REV-2692

PR #25, opened 2022-04-27
feat: Integrated openedx-filters and implemented a small PoC pipeline

inventhouse:
**Description:** Initial proof-of-concept pipeline using [openedx-filters](https://github.com/openedx/openedx-filters)

**JIRA:** [[REV-2704] Hoist Order History: semi-synchronous internal communication mechanism POC & ADR - JIRA](https://openedx.atlassian.net/browse/REV-2704)

**Installation instructions:** 
- Activate commerce-coordinator virtualenv
- `make requirements`

**Testing instructions:**

1. `./manage.py runserver localhost:8140 --settings=commerce_coordinator.settings.
local`
2. Hit http://localhost:8140/demo_lms/demo_get_sample_data/
3. See some sample data
4. Edit pipeline in `commerce_coordinator/settings/base.py` to change the order of the steps, drop a step, etc.

**Author concerns:** 
- ~Still need to write ADR~ Added a draft

- Currently uses openedx-filters directly, more work will be needed to:
  - Add tracing
  - Integrate pipeline configuration with signal configs
  - Add configuration validation

- Currently doesn't demonstrate cross-app connections (I don't think that will be substantially different though)

- openedx-filters are somewhat complex and documentation is ...sparse


commerce_coordinator/apps/demo_lms/pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+# By overriding `run_filter` and adding expected arguments to the signature, we get a small amount checking that the
+# name, for example here "sample_data", was among the arguments passed in.
+class SomeData(PipelineStep):
+    """
+    Adds data to the sample data list.
+    """
+    def run_filter(self, sample_data, *args, **kwargs):  # pylint: disable=arguments-differ
+        return {
+            "sample_data": sample_data + ["A string", {"dict": "like me"}],
+        }
```
dianekaplan, line 21: To confirm, if we were doing the Order History example, we would have an OrderHistory class (here or wherever we were saving our pipeline) and then within its run_filter method that's where we'd call out to oscar or commercetools, is that right?

inventhouse: Correct.

docs/decisions/0004-openedx-filters-for-close-communication.rst
```diff
@@ -0,0 +1,45 @@
...
+Context
+*******
+
+As outlined in `ADR-3 <./0003-internal-communication.rst>`__, Django signals provide excellent decoupling of components and we will be extending them to improve tracing/debugging, validation, and holistic system comprehension.
+
+There are cases for which signals are a poor fit when a component or action requires a direct response, however, we still want the source of that response data to be be decoupled and highly configurable.
+
+There is substantial cost to having multiple distinct but similar mechanisms in an ecosystem; conversely there are benefits to using a common foundation as long as basically serves the need, even if it may not fit as well as a bespoke solution.
+
+- Shared development and maintenence burdon
```
dianekaplan, line 24: little typo: burden

docs/decisions/0004-openedx-filters-for-close-communication.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Context
+*******
+
+As outlined in `ADR-3 <./0003-internal-communication.rst>`__, Django signals provide excellent decoupling of components and we will be extending them to improve tracing/debugging, validation, and holistic system comprehension.
+
+There are cases for which signals are a poor fit when a component or action requires a direct response, however, we still want the source of that response data to be be decoupled and highly configurable.
+
+There is substantial cost to having multiple distinct but similar mechanisms in an ecosystem; conversely there are benefits to using a common foundation as long as basically serves the need, even if it may not fit as well as a bespoke solution.
+
```
dianekaplan, line 23: ^ This sentence was a little confusing for me to parse. If I'm understanding your meaning correctly, I think this may be a simpler restatement:

Signals and Filters are similar-yet-distinct mechanisms. Using both in the same ecosystem (as opposed to only signals) has a potential confusion cost, but it also gives benefit as long as each serves their purpose well.

docs/decisions/0004-openedx-filters-for-close-communication.rst
```diff
@@ -0,0 +1,45 @@
...
+- Shared development and maintenence burdon
+- Skills and knowledge translation
+
+`Openedx-filters <https://github.com/openedx/openedx-filters>`__ is a mechanism being developed to make the edx-platform project more extensible, and specifically to allow configurable functionality that augments, filters, or transforms data at key points in various processes.  It is expressly designed to compliment signals-driven extensions.
+
+Decision
+********
+
+When signals are not feasible for component communication, we will use openedx-filters.
+
```
dianekaplan, line 33: It'll be helpful to specify what makes something not feasible for signals. So far we mentioned asynchronous-ish calls, any other criteria that'll likely be better handled by filters than by signals?

docs/decisions/0004-openedx-filters-for-close-communication.rst
inventhouse: Thanks, I will try to clarify; I mean to make the case that creating our own version of filters would have costs and using an existing implementation would have benefits

docs/decisions/0004-openedx-filters-for-close-communication.rst
inventhouse: Agreed, I can try to better spell out when to use which mechanism; perhaps an example would also be helpful?

docs/decisions/0004-openedx-filters-for-close-communication.rst
inventhouse: Re-ordered a bit and tried to clarify my intention, let me know if it's clearer.

dianekaplan: Yes to the reordering, that makes it easier to understand! I actually was all wrong about what you meant in the last sentence: I thought you were comparing using signals & filters vs signals for everything. I had no idea that your last sentence was referring to signals & filters vs signals plus creating our own version of filters(!). (I thought the 'bespoke' bit was choosing to use both, but you meant actually write it). 

To make sure I'm understanding, is this an accurate restatement of what you're saying?
There is substantial cost to having multiple distinct but similar mechanisms in an ecosystem (both signals and filters); but these provide the benefit of having a common foundation as long as each part serves its need. The alternative would be writing a bespoke version of filters which could fit better for us, but that would add time, effort and complexity. Therefore we're choosing to use signals and filters as they're already defined.

docs/decisions/0004-openedx-filters-for-close-communication.rst
dianekaplan: Sure that sounds great!

docs/decisions/0004-openedx-filters-for-close-communication.rst
inventhouse: Tried to clarify a bit more; the cost I was referring to in this paragraph is _only_ the cost of creating a bespoke solution that is similar to filters but different.

Should I also add a paragraph addressing the cost of having two mechanisms for components to communicate?

dianekaplan: Dude this is so much clearer! love it! And now I think the first parts are so clear that we can/should just remove the last paragraph, right? Like, if we're *not* going to make a bespoke solution, will it be necessary to say anything about not doing it?

inventhouse: Dropped the bit about not building a bespoke solution, but kept the part about benefits of using a shared foundation.

docs/decisions/0004-openedx-filters-for-close-communication.rst
```diff
@@ -0,0 +1,48 @@
...
+    - **Draft** if the decision is still preliminary and in experimental phase
+    - **Accepted** *(date)* once it is agreed upon
+    - **Superseded** *(date)* with a reference to its replacement if a later ADR changes or reverses the decision
+
+Context
+*******
+
+As outlined in `ADR-3 <./0003-internal-communication.rst>`__, Django signals provide excellent decoupling of components and we will be extending them to improve tracing/debugging, validation, and holistic system comprehension.
+
+Signals work well in cases where some action simply needs to be triggered at the appropreate time or a system only needs to be notified of an event but does not need to return any sort of response.
```
pshiu, line 20:
```suggestion
Signals work well in cases where some action simply needs to be triggered at the appropriate time or a system only needs to be notified of an event but does not need to return any sort of response.
```

docs/decisions/0004-openedx-filters-for-close-communication.rst
```diff
@@ -0,0 +1,48 @@
...
+
+There are cases for which signals are a poor fit, such as when a component or action requires a direct response; however, we still want the source of that response data to be be decoupled and highly configurable.
+
+`Openedx-filters <https://github.com/openedx/openedx-filters>`__ is a mechanism being developed to make the edx-platform project more extensible, and specifically to allow configurable functionality that augments, filters, or transforms data at key points in various processes.  It is expressly designed to compliment signals-driven extensions.
+
+There are benefits to using a common foundation as long as basically serves the need, especially shared development and maintenence burden and portability of knowledge and skills among projects in the ecosystem.
```
pshiu, line 26:
```suggestion
There are benefits to using a common foundation as long as it basically serves the need, especially shared development and maintenance burden and portability of knowledge and skills among projects in the ecosystem.
```

commerce_coordinator/settings/base.py
```diff
@@ -251,7 +251,7 @@ def root(*x):
     'commerce_coordinator.apps.core.signals.test_signal': [
         'commerce_coordinator.apps.demo_lms.signals.test_receiver',
         'commerce_coordinator.apps.core.signals.test_receiver_exception',
-        'commerce_coordinator.apps.core.signals.test_celery_task',
+        # 'commerce_coordinator.apps.core.signals.test_celery_task',
```
pshiu, line 254: Does this need to be commented out?

commerce_coordinator/apps/demo_lms/pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+from openedx_filters import PipelineStep
+
+
+# PipelineSteps augment, filter, or transform data passed into them and return a dict that gets merged into the
+# `**kwargs` passed to the next step in the pipeline.  The merge is done using
+# [dict.update()](https://docs.python.org/3/library/stdtypes.html#dict.update)
+#
+# By overriding `run_filter` and adding expected arguments to the signature, we get a small amount checking that the
+# name, for example here "sample_data", was among the arguments passed in.
+class SomeData(PipelineStep):
```
pshiu, line 14: Would it be clearer to rename this class to `AddSomeData`? (And correspondingly, rename `SomeMoreData` to `AddSomeMoreData`?)

commerce_coordinator/apps/demo_lms/views.py
```diff
@@ -83,3 +84,16 @@ def demo_purchase_complete(_):
...
+    """
+    Demonstrate a workflow that gets data from providers configured as pipeline steps.
+    """
+
+    sample_data = SampleDataRequested.run_filter()
+    return JsonResponse({"sample_data": sample_data})
+
+    # BJH: try using run_pipeline directly and see what else we get back
+    # results = SampleDataRequested.run_pipeline(sample_data=[])
+    # return JsonResponse(results)
```
pshiu, line 99: Given [this comment](https://github.com/edx/commerce-coordinator/pull/25/files#diff-f670ac19fec49481dce669f76686f569169a938818112ded8c8a1344d2910c7fR17) in the class definition `SampleDataRequested`, could the above be removed, given it seems we've settled on using `run_filter()` instead of `run_pipeline()`?

commerce_coordinator/settings/base.py
```diff
@@ -270,3 +270,14 @@ def root(*x):
 
 # API URLs
 ECOMMERCE_URL = None
+
+# Filters PoC
+OPEN_EDX_FILTERS_CONFIG = {
```
pshiu, line 275: How will this work once we have multiple Django apps using filters? And especially if two apps need to use the same filter but with different pipeline configurations?

pshiu (approved):
Exciting work!

commerce_coordinator/settings/base.py
inventhouse: I don't believe so, filters are not _really_ designed to be used from more than one place.  Fortunately for our work there's nothing to prevent that, but the filter will run the same pipeline wherever it's used; I'm not completely sure if that's a good thing or a bad thing, but I do think it's a limitation we can live with.

commerce_coordinator/apps/demo_lms/views.py
inventhouse: Ah, yes, this was test code, I shoulda tagged it with my `NOCOMMIT` marker, I'll clean it up.

PR #35, opened 2022-06-24
docs: tidy up autodoc imports for sphinx

pshiu:
## Description

Cleans up autodoc imports for sphinx so it matches https://github.com/openedx/edx-cookiecutters/pull/204.


pshiu:
Won't fix, see issue above.

PR #45, opened 2022-08-16
feat: [REV-2847] add ecommerce/redeem_enrollment_code incoming connection & signals

pshiu:
## Description

We are doing an internal hackathon for some rapid prototyping in the Coordinator. The objective of this hackathon is to get proof of concept code for redeeming enrollment codes (aka 100% coupons) via an existing ecommerce system in our company called Titan.

This PR adds:
* An incoming external connection through a new Coordinator API resource at `/ecommerce/redeem_enrollment_code`
* A internal connection through a new signal called `enrollment_code_redemption_requested_signal` and its receiver `enrollment_code_redemption_requested_create_order`
* A Celery task called `enrollment_code_redemption_requested_create_order_task` that will in the future call the Titan `/redeemEnrollmentCode` endpoint in a future PR in this repo.

## Additional information

* Jira: [REV-2847](https://2u-internal.atlassian.net/browse/REV-2847)
* Previous art: https://github.com/edx/commerce-coordinator/pull/44
* Draft sequence diagram:
    ```mermaid
    sequenceDiagram
        actor user as User
        participant lms as edX LMS
        participant ecommerce as edX Ecommerce
        participant frontend as Enrollment Page Frontend
        participant coordinator as edX Coordinator
        participant titan as Titan
        user->>frontend: GET {enrollment code URL}
        frontend->>+coordinator: POST /redeem_enrollment_code
        coordinator->>-titan: POST /redeemEnrollmentCode
        titan->>+coordinator: GET /checkEnrollmentCode
        coordinator->>+ecommerce: GET /coupons/check/[id]
        ecommerce-->>-coordinator: /coupons/check/[id] response
        coordinator-->>-titan: /checkEnrollmentCode response
        titan->>+coordinator: POST /fulfill
        coordinator->>ecommerce: POST /coupons/invalidate/[id]
        coordinator->>-lms: POST /api/enrollment/v1/enrollment
    ```
* Objective of this PR is to implement `frontend->>+coordinator: POST /redeem_enrollment_code` and add the wiring for `coordinator->>-titan: POST /redeemEnrollmentCode` to happen in the Coordinator.

## Testing instructions

It is a known issue this PR has no tests. Tests will be created after the proof of concept hackathon.

PR #46, opened 2022-08-18
feat: [REV-2847] add TitanAPIClient and TitanOAuthAPIClient

pshiu:
## Description

We are doing an internal hackathon for some rapid prototyping in the Coordinator. The objective of this hackathon is to get proof of concept code for redeeming enrollment codes (aka 100% coupons) via an existing ecommerce system in our company called Titan.

This PR adds:
* Two API clients for Titan, one using OAuth (`TitanOAuthAPIClient`) and another using only an API key (`TitanAPIClient`).
* Wiring to the Celery task `enrollment_code_redemption_requested_create_order_task` to call the Titan `/redeemEnrollmentCode` endpoint using `TitanAPIClient`.
* A new Celery task called `enrollment_code_redemption_requested_create_order_oauth_task` that calls the Titan `/redeemEnrollmentCode` endpoint using `TitanOAuthAPIClient`.

This PR also:
* Renames a couple of parameters from https://github.com/edx/commerce-coordinator/pull/45
* Adds environmental variables for `TITAN_URL` and `TITAN_API_KEY`.
* Removes unnecessary configuration in devstack settings file.

## Additional information

* Jira: [REV-2847](https://2u-internal.atlassian.net/browse/REV-2847)
* Previous art: https://github.com/edx/commerce-coordinator/pull/44
* Draft sequence diagram:
    ```mermaid
    sequenceDiagram
        actor user as User
        participant lms as edX LMS
        participant ecommerce as edX Ecommerce
        participant frontend as Enrollment Page Frontend
        participant coordinator as edX Coordinator
        participant titan as Titan
        user->>frontend: GET {enrollment code URL}
        frontend->>+coordinator: POST /redeem_enrollment_code
        coordinator->>-titan: POST /redeemEnrollmentCode
        titan->>+coordinator: GET /checkEnrollmentCode
        coordinator->>+ecommerce: GET /coupons/check/[id]
        ecommerce-->>-coordinator: /coupons/check/[id] response
        coordinator-->>-titan: /checkEnrollmentCode response
        titan->>+coordinator: POST /fulfill
        coordinator->>ecommerce: POST /coupons/invalidate/[id]
        coordinator->>-lms: POST /api/enrollment/v1/enrollment
    ```
* Objective of this PR is to add the wiring for `coordinator->>-titan: POST /redeemEnrollmentCode` to happen in the Coordinator.

## Testing instructions

It is a known issue this PR has no tests. Tests will be created after the proof of concept hackathon.

### DIFF
--- a/commerce_coordinator/apps/ecommerce/clients.py
+++ b/commerce_coordinator/apps/ecommerce/clients.py
@@ -2,6 +2,7 @@
 API clients for ecommerce app.
 """
 import logging
+from urllib.parse import urljoin
 
 import requests
 from django.conf import settings
@@ -15,7 +16,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
     """
     API client for calls to the edX Ecommerce service.
     """
-    api_base_url = str(settings.ECOMMERCE_URL).strip('/') + '/api/v2'
+    api_base_url = urljoin(settings.ECOMMERCE_URL, '/api/v2')
 
     def get_orders(self, query_params):
         """
@@ -30,7 +31,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
 
         """
         try:
-            resource_url = self.api_base_url + '/orders'
+            resource_url = urljoin(self.api_base_url, '/orders')
             response = self.client.get(resource_url, params=query_params)
             response.raise_for_status()
             return response.json()
--- a/commerce_coordinator/apps/ecommerce/clients.py
+++ b/commerce_coordinator/apps/ecommerce/clients.py
@@ -30,7 +30,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
 
         """
         try:
-            resource_url= self.api_base_url + '/orders'
+            resource_url = self.api_base_url + '/orders'
             response = self.client.get(resource_url, params=query_params)
             response.raise_for_status()
             return response.json()
--- a/commerce_coordinator/apps/ecommerce/clients.py
+++ b/commerce_coordinator/apps/ecommerce/clients.py
@@ -15,7 +15,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
     """
     API client for calls to the edX Ecommerce service.
     """
-    api_base_url = str(settings.ECOMMERCE_URL) + '/api/v2/'
+    api_base_url = str(settings.ECOMMERCE_URL).strip('/') + '/api/v2'
 
     def get_orders(self, query_params):
         """
@@ -30,8 +30,8 @@ class EcommerceApiClient(BaseEdxOAuthClient):
 
         """
         try:
-            endpoint = self.api_base_url + 'orders/'
-            response = self.client.get(endpoint, params=query_params)
+            resource_url= self.api_base_url + '/orders'
+            response = self.client.get(resource_url, params=query_params)
             response.raise_for_status()
             return response.json()
         except requests.exceptions.HTTPError as exc:
--- a/commerce_coordinator/apps/ecommerce/clients.py
+++ b/commerce_coordinator/apps/ecommerce/clients.py
@@ -12,7 +12,7 @@ from commerce_coordinator.apps.core.clients import BaseEdxOAuthClient
 logger = logging.getLogger(__name__)
 
 
-class EcommerceApiClient(BaseEdxOAuthClient):
+class EcommerceAPIClient(BaseEdxOAuthClient):
     """
     API client for calls to the edX Ecommerce service.
     """
--- a/commerce_coordinator/apps/ecommerce/views.py
+++ b/commerce_coordinator/apps/ecommerce/views.py
@@ -1,8 +1,6 @@
 """
 Views for the ecommerce app
 """
-import logging
-
 from edx_rest_framework_extensions.permissions import LoginRedirectIfUnauthenticated
 from rest_framework.response import Response
 from rest_framework.throttling import UserRateThrottle
@@ -11,8 +9,6 @@ from rest_framework.views import APIView
 from commerce_coordinator.apps.core.signal_helpers import format_signal_results
 
 from .signals import enrollment_code_redemption_requested_signal
-
-logger = logging.getLogger(__name__)
 
 
 class RedeemEnrollmentCodeView(APIView):
--- a/commerce_coordinator/apps/ecommerce/views.py
+++ b/commerce_coordinator/apps/ecommerce/views.py
@@ -29,7 +29,7 @@ class RedeemEnrollmentCodeView(APIView):
             Dictionary with results from signal dispatch to redeem an enrollment code.
 
         Raises:
-            PermissionDenied: Djano was unable to determine the user's id or email.
+            PermissionDenied: Djano was unable to determine the user's id, username, or email.
         """
 
         sku = request.query_params.get('sku')
@@ -37,6 +37,8 @@ class RedeemEnrollmentCodeView(APIView):
 
         if not request.user.id:
             raise PermissionDenied(detail="Could not detect user id.")
+        if not request.user.username:
+            raise PermissionDenied(detail="Could not detect username.")
         if not request.user.email:
             raise PermissionDenied(detail="Could not detect user email.")
 
@@ -48,6 +50,7 @@ class RedeemEnrollmentCodeView(APIView):
         results = enrollment_code_redemption_requested_signal.send_robust(
             sender=self.__class__,
             user_id=request.user.id,
+            username=request.user.username,
             email=request.user.email,
             sku=sku,
             coupon_code=code,
--- a/commerce_coordinator/apps/ecommerce/views.py
+++ b/commerce_coordinator/apps/ecommerce/views.py
@@ -27,6 +27,9 @@ class RedeemEnrollmentCodeView(APIView):
 
         Returns:
             Dictionary with results from signal dispatch to redeem an enrollment code.
+
+        Raises:
+            PermissionDenied: Djano was unable to determine the user's id or email.
         """
 
         sku = request.query_params.get('sku')
--- a/commerce_coordinator/apps/ecommerce/views.py
+++ b/commerce_coordinator/apps/ecommerce/views.py
@@ -2,6 +2,7 @@
 Views for the ecommerce app
 """
 from edx_rest_framework_extensions.permissions import LoginRedirectIfUnauthenticated
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.response import Response
 from rest_framework.throttling import UserRateThrottle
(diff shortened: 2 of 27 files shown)


commerce_coordinator/apps/ecommerce/views.py
```diff
@@ -12,30 +11,40 @@
...
+        """
 
-        code = request.query_params.get('code')
         sku = request.query_params.get('sku')
+        code = request.query_params.get('code')
 
-        if not code:
-            return Response({'error': 'Code not provided.'})
+        if not request.user.id or not request.user.email:
+            raise PermissionDenied(detail="Could not detect user id or email.")
```
colinbrash, line 36: Are both required, or just one?  Assuming both are required, consider tweaking the exception detail to be slightly clearer -- at least to me it reads like one is required but neither are present.

commerce_coordinator/apps/titan/clients.py
```diff
@@ -0,0 +1,103 @@
...
+import requests
+from celery.utils.log import get_task_logger
+from django.conf import settings
+from edx_rest_api_client.client import OAuthAPIClient
+
+# Use special Celery logger for tasks client calls.
+logger = get_task_logger(__name__)
+
+
+class TitanAPIClient():
```
colinbrash, line 14: Does edX care about capitalization style?  Here API is all caps, but EcommerceApiClient has it camel case.  Ignore if we don't, just mentioning it in case we do!

commerce_coordinator/apps/titan/clients.py
```diff
@@ -0,0 +1,103 @@
...
+
+        Arguments:
+            resource_path: the path of the API resource to POST to
+            data: the dictionary to send to the API resource
+        Returns:
+            dict: Dictionary represention of JSON returned from API
+
+        """
+        try:
+            resource_url = self.api_base_url.strip('/') + resource_path
```
colinbrash, line 58: See my prior question about this pattern for building urls.  `strip()` here is not technically needed, and is inconsistent with leaving it out on [Line 33](https://github.com/edx/commerce-coordinator/pull/46/files#diff-9e9e16f53a74ae158cfc7a9e805b26c53fb3ea088cbb1cbeb230f3c14b8f3c91R33).  Not a major issue, but it does seem like there could be a better way...

colinbrash (commented):
Wasn't sure if you were just pushing this through quickly.  None of my comments are critical if you just want to get it in for the hackathon work!

colinbrash (approved):

commerce_coordinator/apps/titan/clients.py
pshiu: I was unsure, so I did a little research. Looks like the balance is in favor of all capital letters for acronyms:
* We use an internal library called `edx_rest_api_client`, which provides for us `OAuthAPIClient`
* We derive our views from Django REST Framework's `APIView`
* [PEP-8](https://peps.python.org/pep-0008/#descriptive-naming-styles) also recommends it: "When using acronyms in CapWords, capitalize all the letters of the acronym"

There is definitely a lot of variable usage across our projects though.

commerce_coordinator/apps/titan/clients.py
pshiu: You're right. I should be doing `strip()` where we receive user input and not litter it everywhere in the code. In this case, it should only be when reading in `settings.TITAN_URL`, because historically we've had trouble standardizing trailing slashes in environmental variables.

My original rational was to remind the coder that `api_base_url` has its trailing slash stripped, and thus `resource_path` must begin with a slash. But that is better done with an explicit check.

I'll remove the `strip()` here, assert `resource_path` begins with a slash, and add it to my list of things to add environmental variable validation for.

commerce_coordinator/apps/ecommerce/views.py
pshiu: Good point, will amend.

commerce_coordinator/apps/ecommerce/views.py
```diff
@@ -12,30 +11,40 @@
...
+        """
+        Redeem an enrollment code for an authenticated user.
+
+        Args:
+            sku: ecommerce partner_sku to be redeemed
+            code: enrollment code (aka, 100 percent off coupon code) to redeem.
+
+        Returns:
+            Dictionary with results from signal dispatch to redeem an enrollment code.
+        """
```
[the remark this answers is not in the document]
pshiu: Yes it should! Thanks for the catch.

commerce_coordinator/apps/ecommerce/clients.py
```diff
@@ -15,7 +15,7 @@ class EcommerceApiClient(BaseEdxOAuthClient):
     """
     API client for calls to the edX Ecommerce service.
     """
-    api_base_url = str(settings.ECOMMERCE_URL) + '/api/v2/'
+    api_base_url = str(settings.ECOMMERCE_URL).strip('/') + '/api/v2'
```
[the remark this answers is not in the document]
pshiu: Maybe I should have just used `urllib.parse.urljoin` then:
```
% python
Python 3.8.11 (default, Aug 23 2021, 18:14:59) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from urllib.parse import urljoin
>>> urljoin("https://example.com", "/test")
'https://example.com/test'
>>> urljoin("https://example.com/", "test")
'https://example.com/test'
>>> urljoin("https://example.com", "test")
'https://example.com/test'
>>> urljoin("https://example.com/", "/test")
'https://example.com/test'
>>> urljoin("https://example.com", "/v1/test")
'https://example.com/v1/test'
>>> urljoin("https://example.com", "/v1/test/")
'https://example.com/v1/test/'
>>> urljoin("https://example.com/", "/v1/test/")
'https://example.com/v1/test/'
>>> urljoin("replace-me", "/v1/test/")
'/v1/test/'
```

Last one will redirect to localhost for our default values in settings/base.py, which isn't the worst I suppose. Will modify.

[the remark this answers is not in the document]
pshiu: Side note: I think the `str()` is in there mostly for when settings.ECOMMERCE_URL is `None`. But `urljoin` takes care of that. It won't handle if settings.ECOMMERCE_URL is a number or multi-value data structure, but it feels safe to not handle those cases in favor of simplicity:

```
>>> from urllib.parse import urljoin
>>> urljoin(None, "/v1/test/")
'/v1/test/'
>>> urljoin(True, "/v1/test/")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/pshiu/.pyenv/versions/3.8.11/lib/python3.8/urllib/parse.py", line 522, in urljoin
    base, url, _coerce_result = _coerce_args(base, url)
  File "/Users/pshiu/.pyenv/versions/3.8.11/lib/python3.8/urllib/parse.py", line 124, in _coerce_args
    raise TypeError("Cannot mix str and non-str arguments")
TypeError: Cannot mix str and non-str arguments
```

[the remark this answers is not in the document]
pshiu: Done in https://github.com/edx/commerce-coordinator/pull/46/commits/d7bd1057c2260f7005fdf1a161244a60e6494cc9.

commerce_coordinator/apps/titan/clients.py
```diff
@@ -0,0 +1,104 @@
...
+        """URL of API service."""
+        return urljoin(settings.TITAN_URL, '/v1')
+
+    @property
+    def api_key_header(self):
+        """Header to add to all API requests to authenticate to endpoint."""
+        return {'X-API-Key': settings.TITAN_API_KEY}
+
+    @property
+    def normal_timeout(self):
```
julianajlk, line 36: Can you move this to `core` so we can use it in other apps?

commerce_coordinator/apps/titan/clients.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @property
+    def api_base_url(self):
+        """URL of API service."""
+        return urljoin(settings.TITAN_URL, '/v1')
+
+    @property
+    def api_key_header(self):
+        """Header to add to all API requests to authenticate to endpoint."""
+        return {'X-API-Key': settings.TITAN_API_KEY}
```
julianajlk, line 33: Do you specify the `Content-Type` of the header somewhere?

pshiu: No, but it's autodetected I think. Here's a sample response, see the `Content-Type` at the end.

```
[2022-09-06 12:58:26,612: DEBUG/ForkPoolWorker-8] commerce_coordinator.apps.titan.tasks.enrollment_code_redemption_requested_create_order_task[459b0255-a169-413e-a419-a012ce11fd82]: Request headers: {'User-Agent': 'python-requests/2.28.1', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'X-API-Key': 'replace-me', 'Content-Length': '114', 'Content-Type': 'application/json'}
```

commerce_coordinator/apps/titan/clients.py
pshiu: Done: https://github.com/edx/commerce-coordinator/pull/46/commits/84d8fc879d846c93dbd05b2baa63912486ef09f2

commerce_coordinator/apps/titan/clients.py
julianajlk: `Header to add to all API requests` - it would be to all API requests to Titan only right?

commerce_coordinator/apps/ecommerce/views.py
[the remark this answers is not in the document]
pshiu: Done: https://github.com/edx/commerce-coordinator/commit/801340011a87b8c31e8f486823808952f92b18c2

commerce_coordinator/apps/ecommerce/views.py
pshiu: Done: https://github.com/edx/commerce-coordinator/commit/76ba447a34ceedbf2c97d8523656f2ae8bb4f9a0

commerce_coordinator/apps/titan/clients.py
pshiu: Done in https://github.com/edx/commerce-coordinator/commit/d7bd1057c2260f7005fdf1a161244a60e6494cc9.

commerce_coordinator/apps/titan/clients.py
pshiu: Done: https://github.com/edx/commerce-coordinator/pull/46/commits/6c4cf3f3f6682371c8554f3533ac4dcabcf5f2a5

PR #47, opened 2022-08-19
feat: Add OrderFulfill endpoint and LMS enrollment signal

julianajlk:
[REV-2847](https://2u-internal.atlassian.net/browse/REV-2847).

Hackathon work for the Theseus Project POC for enrollment codes - fulfill, which is separated into 3 parts:
1) Adding a `/order/fulfill` endpoint that Titan can make a `POST` request to for order fulfillment after the purchase is approved, plus a Celery task for calling LMS enrollment.
2) `POST` request to LMS enrollment to enroll the user.
3) Coupon invalidation to mark it as used.

Screen Shot 2022-08-28 at 11 27 41 AM

**This PR is part 1:**
- Adds `OrderFulfill` API to Titan app, including a serializer file (`POST`)
- Adds Titan -> LMS signal and Celery task for order fulfillment

**Testing instructions:**

1. Checkout this branch `julianajlk/REV-2847/fulfill-titan`
2. Follow instructions on the README of this repo to start redis in `devstack` and start Celery in `commerce-coordinator` 
3. Go to http://localhost:8140/titan/fulfill/
4. Insert data
4. You should get a 200 OK POST response (response should be "" for now, error: false) and on the Celery task side, you should get a log with 
`Task commerce_coordinator.apps.lms.tasks.fulfill_order_placed_send_enroll_in_course_task [13604e04-1f01-4ebd-8ba1-f1188e891dbc] succeeded`


commerce_coordinator/apps/titan/views.py
```diff
@@ -2,6 +2,81 @@
...
+        # deny global queries
+        if not request.user.username:
+            raise PermissionDenied(detail="Could not detect username.")
+
+        titan_order_id = request.data.get('titan_order_id')
+        logger.info('Fulfillment requested for order id [%s].', titan_order_id)
+        serializer = OrderFulfillSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            import pdb; pdb.set_trace()
```
julianajlk, line 80: Will remove in the next commit

commerce_coordinator/apps/titan/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+    course_id = serializers.CharField(max_length=200)
+    date_placed = serializers.CharField(max_length=200)
+    edx_lms_user_id = serializers.IntegerField()
+    mode = serializers.CharField(max_length=200)
+    partner_sku = serializers.CharField(max_length=200)
+    titan_order_uuid = serializers.IntegerField()
+    edx_lms_username = serializers.CharField(max_length=200)
+
+    def create(self, validated_data):
+        # TODO: perform checks here
```
pshiu, line 18: Do more checks need to be performed here?

commerce_coordinator/apps/titan/serializers.py
```diff
@@ -0,0 +1,26 @@
+""" Serializers for data manipulated by commerce-coordinator titan API endpoints. """
+
+from rest_framework import serializers
+
+
+class OrderFulfillSerializer(serializers.Serializer):
```
pshiu, line 6: Instead of having a create/update, what do you think about https://www.django-rest-framework.org/api-guide/serializers/#overriding-save-directly?

commerce_coordinator/apps/titan/views.py
```diff
@@ -2,6 +2,96 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class OrderFulfillView(APIView):
+    """
+    API for order fulfillment that is called from Titan.
+    """
+    permission_classes = [LoginRedirectIfUnauthenticated]
+    authentication_classes = [SessionAuthentication, BasicAuthentication]
```
pshiu, line 27: I think we can remove `authentication_classes`. It's defined upstream.

commerce_coordinator/apps/titan/views.py
```diff
@@ -2,6 +2,96 @@
...
+        edx_lms_username = request.data.get('edx_lms_username')
+
+        # TODO: add enterprise data for enrollment API here
+
+        # TODO: add credit_provider data here
+        # /ecommerce/extensions/fulfillment/modules.py#L315
+
+        # deny global queries
+        if not request.user.username:
+            raise PermissionDenied(detail="Could not detect username.")
```
pshiu, line 75: `request.user.username` is the service account Titan is using. We should check that using `permission_classes`.

commerce_coordinator/apps/titan/views.py
```diff
@@ -2,6 +2,96 @@
...
+
+        logger.info(
```
pshiu, line 77: Note that we already log at the signal level. But if you want to log here as well, may want to consider adding more parameters.

commerce_coordinator/apps/titan/views.py
```diff
@@ -2,6 +2,96 @@
...
+from .signals import fulfill_order_placed_signal
+
+logger = logging.getLogger(__name__)
+
+
+class OrderFulfillView(APIView):
+    """
+    API for order fulfillment that is called from Titan.
+    """
+    permission_classes = [LoginRedirectIfUnauthenticated]
```
pshiu, line 26: Probably need to restrict to backend services only. Let's use [`IsAdminUser`](https://www.django-rest-framework.org/api-guide/permissions/#isadminuser) for now but we'll probably have to re-visit this later.

pshiu (approved):
Looks great - thank you!

commerce_coordinator/apps/lms/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+    mode,
+    partner_sku,
+    titan_order_uuid,
+    edx_lms_username,
+):
+    """
+    Celery task for order placed fulfillment and enrollment via LMS Enrollment API.
+    """
+    logger.info(
+        f'LMS fulfill_order_placed_send_enroll_in_course_task fired with user {edx_lms_user_id},'
```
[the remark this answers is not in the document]
julianajlk: Sure, it's in alphabetical order in L13-20 but the variable name is not always how we'll refer to it in the log (ie. LMS user ID vs `edx_lms_user_id`)

commerce_coordinator/apps/titan/views.py
julianajlk: Gotcha, will remove from LMS app views as well

commerce_coordinator/apps/titan/views.py
julianajlk: Removed. Your "deny global queries" makes more sense now that I know the context of order history for the user

commerce_coordinator/apps/titan/views.py
julianajlk: I don't necessarily want to add everything due to that reason, but wanted to indicate that it started the fulfill process

julianajlk: I would just log the order ID, but since order ID in this case doesn't matter to us since it's the Titan Order ID, I added some other info but not all.

commerce_coordinator/apps/titan/serializers.py
julianajlk: Nice find! I prefer this since we don't need an `update()`. Thank you!

commerce_coordinator/apps/titan/serializers.py
julianajlk: Modifying to `save()` as per below comment

commerce_coordinator/apps/titan/serializers.py
julianajlk: Decided to remove the serializer assuming all of the data validation will be done on the `edx-platform` or `ecommerce` endpoint, and `commerce-coordinator` is just the messenger/carrier of data

PR #57, opened 2022-10-31
Updates for Docker Provisioning, And first attempt at Titan support

grmartin:
# Updates for Docker Provisioning, As well as first attempt at Titan support.

- `provision-commerce-coordinator.sh` now creates the full Docker stack.
- `docker-compose.yml` modified to support devstack and titan networks.
- `commerce_coordinator/settings/devstack.py` now points to the default instance name for the titan application (though im unsure if the path is correct), can be configured via EnvVar
- `Makefile` changes for provisioning as well as manual orchestration of the 3 stacks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,12 @@ validate_translations: fake_translations detect_changed_source_translations ## i
 docker_build:
 	docker build . -f Dockerfile -t openedx/commerce-coordinator
 
+dev.provision_docker: # start LMS, and Setup a clean commerce-coordinator stack. You will need to start titan yourself.
+	bash provision-commerce-coordinator.sh
+
+dev.run_test_query:
+	curl  http://localhost:8140/demo_lms/test/
+
 # devstack-themed shortcuts
 dev.up: # Starts all containers
 	docker-compose up -d
@@ -161,6 +167,21 @@ dev.down: # Kills containers and all of their data that isn't in volumes
 
 dev.stop: # Stops containers so they can be restarted
 	docker-compose stop
+
+dev.multistack.up:
+	docker compose -p titan up
+	bash find-start-lms.sh
+	docker-compose up -d --build
+
+dev.multistack.stop:
+	docker compose -p titan stop
+	docker compose -p devstack stop
+	docker compose stop
+
+dev.multistack.start:
+	docker compose -p titan start
+	bash find-start-lms.sh
+	docker compose start
 
 app-shell: # Run the app shell as root
 	docker exec -u 0 -it commerce-coordinator.app bash
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,9 @@ Initial Setup: steps and confirmation
 
 Note: this setup process is temporary; we will be working on a Tutor plugin
 
+Local
+======
+
 .. code-block::
 
   # Clone the repository
@@ -127,6 +130,17 @@ Local testing with Celery
   "<function demo_purchase_complete_enroll_in_course at 0x10e18a670>": ""
  }
 
+With Docker
+===========
+
+1. First start Titan (if you have it)
+2. Execute `make dev.provision_docker`
+
+> Note: This will attempt to connect to LMS and create the required superusers, please ensure you have the edX devstack setup first.
+
+After you can manage the stack by calling `make dev.up`, `make dev.down` (delete) or `make dev.stop`.
+
+As of the time of this writing, you must have run `make dev.up.ecommerce+lms+redis` in edX's devstack as a prerequisite to this one.
 
 License
 *******
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,24 +1,20 @@
 version: "2.1"
 services:
   db:
-    image: mysql:5.6
+    image: edxops/mysql:5.7
     container_name: commerce-coordinator.db
     environment:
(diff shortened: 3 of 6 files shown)


find-start-lms.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+RED="\x1B[31m"
+GREEN="\x1B[32m"
+NC="\x1B[0m"
+
+if [ -z ${DEVSTACK_WORKSPACE+x} ]; then DEVSTACK_WORKSPACE=$(pwd)/../../openedx; fi
```
pshiu, line 7: I think the default setup is to have the clone of the devstack repository be a sibling directory to commerce-coordinator. See [these devstack instructions](https://github.com/openedx/devstack#directions-to-setup-devstack).

pshiu, line 7: Doesn't the `+x` substitute `DEVSTACK_WORKSPACE` with `x` if it's not there? May want to be careful about changing `DEVSTACK_WORKSPACE` in general because devstack uses it a lot internally.

Makefile
```diff
@@ -157,11 +163,32 @@ dev.up.build:
 	docker-compose up -d --build
 
 dev.down: # Kills containers and all of their data that isn't in volumes
-	docker-compose down
+	docker-compose down -v
```
pshiu, line 166: I'm a little worried because edX developers are used to the idea that `dev.down` is not destructive. Is there a reason we need to kill volumes here?

Makefile
```diff
@@ -157,11 +163,32 @@ dev.up.build:
...
 
 dev.stop: # Stops containers so they can be restarted
 	docker-compose stop
 
+dev.multistack.up:
+	docker compose -p titan up
+	bash find-start-lms.sh
+	docker-compose up -d --build
+
+## On second thoughtm since this obliterates everything... this is dangerous.
```
pshiu, line 176: Yes! Very dangerous. Highly agree.

Makefile
grmartin: ok, np.

README.rst
```diff
@@ -31,6 +31,9 @@ Initial Setup: steps and confirmation
 
 Note: this setup process is temporary; we will be working on a Tutor plugin
 
+Local
+======
```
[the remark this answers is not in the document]
grmartin: woo. thanks. i dont know if this is general or specific to this line, but in this case i was using the same heading format as the rest of the document :)

docker-compose.yml
```diff
@@ -27,14 +25,19 @@ services:
...
     ports:
       - "8140:8140"
-    networks:
-      - devstack_default
     stdin_open: true
     tty: true
+    networks:
+      - devstack_default # edX Dev Stack
+      - titan_default    # GetSmarter Titan
+      - default          # Just these containers
```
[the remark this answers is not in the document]
grmartin: so the database and memcache aret in the titan or devstack networks but the application server can see everything it needs without bleeding in to other networks too much.

find-start-lms.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
grmartin: `depends_on` cant reach in to other docker compose files unless they are included by file... and if they are in the same directory... compose's executable confuses PWD/CWD contexting... 

So if im in `a`.... `
a$ docker compose -f ./docker-compose.yml -f ../b/docker-compose.yml up`... the docker compose config in `b` will be looking in a for all its includes, settings, dockerfiles, etc in `a`, instead of its own containing folder.

find-start-lms.sh
grmartin: i was grouping by github org, heh... seemed more rational, will fix.

grmartin: i am aware, but it is pointing to the path correctly... as for `+x` i think shellcheck made me change it to this... ill look in to it again.

grmartin: `if [ -z ${var+x} ]; then echo "var is unset"; else echo "var is set to '$var'"; fi`

where ${var+x} is a [parameter expansion](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) which evaluates to nothing if var is unset, and substitutes the string x otherwise.

From: https://stackoverflow.com/a/13864829

Do you have a method youd prefer to use for this @pshiu? Otherwise i think im ready to merge

grmartin:
Also @pshiu  any idea on this? looks like a bad signal, not something i did: https://github.com/edx/commerce-coordinator/actions/runs/3574923864/jobs/6010963296

pshiu (approved):
LGTM but please see comments.

find-start-lms.sh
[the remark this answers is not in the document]
grmartin: Sure, but ill do so on Line 2 ;)

find-start-lms.sh
grmartin: Yes, as `if [ -z ${DEVSTACK_WORKSPACE+x} ]; then DEVSTACK_WORKSPACE=$(pwd)/..; fi` this is correct, right?

grmartin:
@pshiu commitlint is broken, any thoughts?

PR #62, opened 2023-02-27
build: add 2u/project-theseus branch to run ci

wdrussell2015:
[REV-3283](https://2u-internal.atlassian.net/browse/REV-3283)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Python CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, 2u/project-theseus]
   pull_request:
     branches:
     - '**'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: Python CI
 on:
   push:
     branches: 
-    - main2u
+    - main
     - 2u/project-theseus
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -2,7 +2,9 @@ name: Python CI
 
 on:
   push:
-    branches: [main]
+    branches: 
+    - main2u
+    - project-theseus
```
pshiu, line 7: Typo

.github/workflows/docker-publish.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - 2u/project-theseus
```
pshiu, line 7: Do we want to publish to Dockerhub for `2u/project-theseus`?

Asking because I thought no, but want to make sure I didn't miss a reason. From my recollections, we publish to Dockerhub to deploy to stage/prod for our k8s setup, and not sure we want to do that yet.

pshiu (changes requested):

.github/workflows/docker-publish.yml
wdrussell2015: Removed, wasnt sure if it was necessary so I had added it there.

.github/workflows/ci.yml
wdrussell2015: fixed

pshiu (approved):

PR #82, opened 2023-06-09
feat: implemented TitanAPIClient.create_payment.

HammadAhmadWaqas:
[THES-190](https://2u-internal.atlassian.net/browse/THES-190)

Implements [Create Payment (commerce-coordinator → titan)](https://github.com/edx/commerce-coordinator-private/blob/a740ebf00ffee93efac0a61c7ed47293f21511b6/docs/api-spec/theseus-api-spec.md#action-create-payment).

### DIFF
--- a/commerce_coordinator/apps/core/constants.py
+++ b/commerce_coordinator/apps/core/constants.py
@@ -17,3 +17,7 @@ class PaymentState(Enum):
     FAILED = 'failed'
     PENDING = 'pending'
     PROCESSING = 'processing'
+
+
+class PaymentMethod(Enum):
+    STRIPE = 'Stripe No Capture'
--- a/commerce_coordinator/apps/titan/clients.py
+++ b/commerce_coordinator/apps/titan/clients.py
@@ -223,6 +223,59 @@ class TitanAPIClient(Client):
 
         return response['data']['attributes']
 
+    def create_payment(
+        self,
+        order_uuid,
+        response_code,
+        payment_method_name,
+        provider_response_body,
+        reference=None,
+        amount=None,
+        payment_date=None,
+        source=None,
+    ):
+        """
+        Request Titan to Create payment.
+
+        Args:
+            order_uuid(str): Order UUID related to this order.
+            response_code(str): Payment attempt response code (payment intent id) provided by stripe.
+            payment_method_name(str): The name of the payment method used for this payment. See enums for valid values.
+            provider_response_body(str): The response JSON dump from a request to the payment provider..
+            reference(str): Optional. Reference to be saved against the payment.
+            amount(float): Optional. Payment amount to apply to order.
+            payment_date(ISO-8601 str): Optional. Payment date to save against the payment.
+            source(dict): Optional. Payment source details to be saved. It should have 'ccType',
+                'gatewayCustomerProfileId', 'gatewayPaymentProfileId', 'lastDigits', 'month', 'name' and 'year'
+                keys in 'str' format.
+        """
+        payload_attributes = {
+            'orderUuid': order_uuid,
+            'responseCode': response_code,
+            'paymentMethodName': payment_method_name,
+            'providerResponseBody': provider_response_body,
+        }
+        if reference is not None:
+            payload_attributes['reference'] = reference
+        if amount is not None:
+            payload_attributes['amount'] = amount
+        if payment_date is not None:
+            payload_attributes['paymentDate'] = payment_date
+        if source is not None:
+            payload_attributes['source'] = source
+
+        response = self._request(
+            request_method='POST',
+            resource_path='payments',
+            json={
+                'data': {
+                    'attributes': payload_attributes
+                }
+            },
+        )
+
+        return response['data']['attributes']
+
     def redeem_enrollment_code(self, sku, coupon_code, user_id, username, email):
         """
         Request Titan to redeem an enrollment code
--- a/commerce_coordinator/apps/titan/tests/test_clients.py
+++ b/commerce_coordinator/apps/titan/tests/test_clients.py
@@ -4,7 +4,7 @@ from unittest.mock import MagicMock, patch
 from django.test import override_settings
 from requests.exceptions import HTTPError
 
-from commerce_coordinator.apps.core.constants import PaymentState
+from commerce_coordinator.apps.core.constants import PaymentMethod, PaymentState
 from commerce_coordinator.apps.core.tests.utils import CoordinatorClientTestCase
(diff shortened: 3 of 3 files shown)


commerce_coordinator/apps/core/constants.py
```diff
@@ -17,3 +17,7 @@ class PaymentState(Enum):
     FAILED = 'failed'
     PENDING = 'pending'
     PROCESSING = 'processing'
+
+
+class PaymentMethod(Enum):
+    STRIPE = 'stripe'
```
HammadAhmadWaqas, line 23: @RylanMorrison 

In the API spec [here](https://github.com/edx/commerce-coordinator-private/blob/main/docs/api-spec/theseus-api-spec.md#enums-1), There is only one enum value for `PaymentMethodName` i.e `Stripe No Capture`. Is there any specific reason to name it `Stripe No Capture`? Do we support any other value for this Enum?

commerce_coordinator/apps/titan/clients.py
```diff
@@ -223,6 +223,59 @@ def update_payment(self, payment_number, payment_state, response_code):
...
+        Request Titan to Create payment.
+
+        Args:
+            order_uuid(str): Order UUID related to this order.
+            response_code(str): Payment attempt response code (payment intent id) provided by stripe.
+            payment_method_name(str): The name of the payment method used for this payment. See enums for valid values.
+            provider_response_body(str): The response JSON dump from a request to the payment provider..
+            reference(str): Optional. Reference to be saved against the payment.
+            amount(int): Optional. Payment amount to apply to order.
+            payment_date(POSIX date): Optional. Payment date to save against the payment.
```
HammadAhmadWaqas, line 247: What should the format be for `paymentDate` field? ISO-8601 or POSIX?

commerce_coordinator/apps/titan/clients.py
```diff
@@ -223,6 +223,59 @@ def update_payment(self, payment_number, payment_state, response_code):
...
+        source=None,
+    ):
+        """
+        Request Titan to Create payment.
+
+        Args:
+            order_uuid(str): Order UUID related to this order.
+            response_code(str): Payment attempt response code (payment intent id) provided by stripe.
+            payment_method_name(str): The name of the payment method used for this payment. See enums for valid values.
+            provider_response_body(str): The response JSON dump from a request to the payment provider..
```
HammadAhmadWaqas, line 244: as this will be a JSON dump, I expect it to be a long str. Is there any cap on `providerResponseBody` Field

HammadAhmadWaqas (commented):
@RylanMorrison few comments for you. Please help to understand.

commerce_coordinator/apps/titan/clients.py
```diff
@@ -223,6 +223,59 @@ def update_payment(self, payment_number, payment_state, response_code):
...
+            reference(str): Optional. Reference to be saved against the payment.
+            amount(int): Optional. Payment amount to apply to order.
```
pshiu, line 246: [upstream spec question] I just noticed `amount` is an integer. Does this mean this is a number of cents?

commerce_coordinator/apps/titan/clients.py
```diff
@@ -223,6 +223,59 @@ def update_payment(self, payment_number, payment_state, response_code):
...
+    ):
+        """
+        Request Titan to Create payment.
+
+        Args:
+            order_uuid(str): Order UUID related to this order.
+            response_code(str): Payment attempt response code (payment intent id) provided by stripe.
+            payment_method_name(str): The name of the payment method used for this payment. See enums for valid values.
+            provider_response_body(str): The response JSON dump from a request to the payment provider..
+            reference(str): Optional. Reference to be saved against the payment.
```
pshiu, line 245: [upstream spec question] I'm not sure what the difference will be between `reference` and `response_code`.

pshiu (approved):
Thanks, Hammad!

commerce_coordinator/apps/core/constants.py
RylanMorrison: We originally went with `Stripe No Capture` to differentiate it from the Stripe payment method we use on the frontend and because we don't want Spree to attempt to process the payment Titan receives (ie send it to Stripe). It turns out flagging the payment method as `no capture` isn't enough to prevent Spree from attempting to process it and I'm currently looking at resolving this in [THES-159](https://2u-internal.atlassian.net/browse/THES-159)

This means this value will change, likely to something like `edX Stripe`, and that is the only value we would support. Maybe there would be other values when we get to supporting other payment providers

commerce_coordinator/apps/titan/clients.py
RylanMorrison: No cap on that field, other than the default cap Postgres puts on varchar database fields which seems to be `10485760` characters, should be enough?

commerce_coordinator/apps/titan/clients.py
RylanMorrison: Currently it's `ISO-8601`. If it were `POSIX` would we need to add timezone to the payloads?

commerce_coordinator/apps/titan/clients.py
RylanMorrison: Reference is an arbitrary string used internally (by our Finance team) for reconciliation as far as I know. Response code is the intent client key from Stripe

commerce_coordinator/apps/titan/clients.py
RylanMorrison: Ah nice catch thanks, will fix that, should be a float

commerce_coordinator/apps/titan/clients.py
HammadAhmadWaqas: Yes, Thank you for specifying it. `ISO-8601` looks good.

commerce_coordinator/apps/titan/clients.py
HammadAhmadWaqas: Yes, This sounds more than enough.

commerce_coordinator/apps/titan/clients.py
HammadAhmadWaqas: Good. I am updating this type to float. (We need to update the spec accordling @RylanMorrison)

commerce_coordinator/apps/titan/clients.py
HammadAhmadWaqas: So this field will be auto-generated by Titan's system? Commerce Coordinator do not need to send it to Titan in their create payment API call?

PR #95, opened 2023-07-07
chore: `pkg-config` added to Docker File

grmartin:
This is to future proof the update of mysql client in future releases. See: https://github.com/edx/edx-arch-experiments/issues/349


Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN apt-get update && apt-get -qy install --no-install-recommends \
  python3.8 \
  python3-pip \
  libmysqlclient-dev \
+ pkg-config \
```
pshiu, line 30: Don't forget to add a comment describing what pkg-config is for above the `RUN` line (see comment on line 23).

pshiu (approved):
Thanks, Glenn. Tests are failing because of an invalid requirements/ci.txt. Running `make upgrade` might fix it.

grmartin:
> make upgrade

```
    wheel_cache = WheelCache(self._cache_dir, self.options.format_control)
TypeError: WheelCache.__init__() takes 2 positional arguments but 3 were given
```

Based on things i found online, this has since been modified. Pinning pip-tools.

PR #96, opened 2023-07-07
feat: Add Pipeline step for UpdateBillingAddress

wdrussell2015:
[THES-201](https://2u-internal.atlassian.net/browse/THES-201)
Adds the Update Billing Address Pipeline step

### DIFF
--- a/commerce_coordinator/apps/titan/pipeline.py
+++ b/commerce_coordinator/apps/titan/pipeline.py
@@ -165,4 +165,4 @@ class UpdateBillingAddress(PipelineStep):
             raise APIException("Error updating the order's billing address details in titan") from exc
         update_billing_address_output = BillingAddressSerializer(data=response)
         update_billing_address_output.is_valid(raise_exception=True)
-        return update_billing_address_output.data
+        return {'billing_address_data': update_billing_address_output.data}
--- a/commerce_coordinator/apps/titan/pipeline.py
+++ b/commerce_coordinator/apps/titan/pipeline.py
@@ -99,7 +99,7 @@ class CreateDraftPayment(PipelineStep):
 
             )
         except HTTPError as exc:
-            logger.exception('[CreateTitanPayment] Failed to create pyment for order_uuid: %s', order_uuid)
+            logger.exception('[CreateTitanPayment] Failed to create payment for order_uuid: %s', order_uuid)
             raise APIException("Error while creating payment on titan's system") from exc
 
         payment_serializer = PaymentSerializer(data=payment)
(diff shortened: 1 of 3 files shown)


commerce_coordinator/apps/titan/pipeline.py
```diff
@@ -140,3 +144,25 @@ def run_filter(self, edx_lms_user_id):  # pylint: disable=arguments-differ
...
+
+        except HTTPError as exc:
+            logger.exception(
+                "[UpdateBillingAddress] Failed to update the billing address for the specified order: %s",
+                order_uuid
+            )
+            raise APIException("Error updating the order's billing address details in titan") from exc
+        update_billing_address_output = BillingAddressSerializer(data=response)
+        update_billing_address_output.is_valid(raise_exception=True)
+        return update_billing_address_output.data
```
HammadAhmadWaqas, line 168: Pipeline steps should return info in wrapped form. for example, here we can return 
```
{ 'billing_address_data': update_billing_address_output.data}
```
This helps to wire these steps in series and pass data to the next step in good shape.

grmartin: i think this is fair, especially as we try to funnel data through multiple steps... Should we also merge this with the input so the next step has everything?

HammadAhmadWaqas: @grmartin Yes, good point. about input, the filter passes inputs to every step in the pipeline, so we do not need to merge input with output.

wdrussell2015: updated to wrap return data

commerce_coordinator/apps/titan/pipeline.py
```diff
@@ -140,3 +144,25 @@ def run_filter(self, edx_lms_user_id):  # pylint: disable=arguments-differ
...
+        api_client = TitanAPIClient()
+        try:
+            response = api_client.update_billing_address(order_uuid, **kwargs)
+
+        except HTTPError as exc:
+            logger.exception(
+                "[UpdateBillingAddress] Failed to update the billing address for the specified order: %s",
+                order_uuid
+            )
+            raise APIException("Error updating the order's billing address details in titan") from exc
```
pshiu, line 165: @wdrussell2015, could you file a ticket to incorporate error handling for titan/pipeline.py? There are a lot of errors from the API spec that we collectively as a team haven't implemented in this file.

For example, I don't think right now we would know the difference between, e.g. an `orderHasCompletedPayments` and an `orderNotFound` error on this endpoint (see [api spec](https://github.com/edx/commerce-coordinator-private/blob/a740ebf00ffee93efac0a61c7ed47293f21511b6/docs/api-spec/theseus-api-spec.md#action-update-billing-address)). We should make sure this occurrence makes it into the logs, at the very least.

(I was going to suggest adding this exception into exception.py to follow suit from some of the other pipeline steps. But it looks like you don't have enough information from the client in this scope to be able to name what exception is happening properly. This work will also have to be done in the ticket above.)

commerce_coordinator/settings/base.py
```diff
@@ -394,6 +394,12 @@ def root(*path_fragments):
             'commerce_coordinator.apps.titan.pipeline.GetTitanActiveOrder',
         ]
     },
+    "org.edx.coordinator.frontend_app_payment.payment.processing.requested.v1": {
+        "fail_silently": False,  # TODO: Coordinator filters should NEVER be allowed to fail silently
+        "pipeline": [  # TODO: Add remaining pipeline steps from THES-202, THES-203, THES-204
```
pshiu, line 399: Thank you for this TODO!

pshiu (approved):
Thank you for this work, Wes.

commerce_coordinator/apps/titan/pipeline.py
wdrussell2015: https://2u-internal.atlassian.net/browse/THES-223

PR #104, opened 2023-08-03
feat: handle sad path for capture context payment mfe

wdrussell2015:
[THES-227](https://2u-internal.atlassian.net/browse/THES-227)


HammadAhmadWaqas (approved):
Thank you.

PR #107, opened 2023-08-10
feat: add create_payment call after a payment failure

wdrussell2015:
[THES-231](https://2u-internal.atlassian.net/browse/THES-231)

Added create payment call to payment_process_save_task and updated tests

Note: Added as a client call but if theres a way to call a pipeline step from within a signal task it might be better to swap to calling the pipeline step instead

### DIFF
--- a/commerce_coordinator/apps/titan/exceptions.py
+++ b/commerce_coordinator/apps/titan/exceptions.py
@@ -21,6 +21,12 @@ class ProcessingAlreadyRequested(APIException):
     default_code = 'already_processing'
 
 
+class StaleOrder(APIException):
+    status_code = 409
+    default_detail = 'The supplied orderUuid does not match the latest active orderUuid for the supplied edxLmsUserId'
+    default_code = 'stale_order'
+
+
 class AlreadyPaid(APIException):
     status_code = 409
     default_detail = 'The payment was already paid.'
--- a/commerce_coordinator/apps/titan/tasks.py
+++ b/commerce_coordinator/apps/titan/tasks.py
@@ -12,9 +12,8 @@ from commerce_coordinator.apps.core.cache import (
     get_paid_payment_state_cache_key,
     get_processing_payment_state_cache_key
 )
-from commerce_coordinator.apps.core.constants import PaymentState
-
-from .clients import TitanAPIClient
+from commerce_coordinator.apps.core.constants import PaymentMethod, PaymentState
+from commerce_coordinator.apps.titan.clients import TitanAPIClient
 
 logger = get_task_logger(__name__)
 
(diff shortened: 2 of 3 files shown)


commerce_coordinator/apps/titan/tasks.py
```diff
@@ -102,8 +101,22 @@ def payment_processed_save_task(
             payment_state_paid_cache_key = get_paid_payment_state_cache_key(payment_number)
             TieredCache.set_all_tiers(payment_state_paid_cache_key, payment, settings.DEFAULT_TIMEOUT)
         elif payment_state == PaymentState.FAILED.value:
+            if 'metadata' in provider_response_body:
+                if 'payment_number' in provider_response_body['metadata']:
+                    provider_response_body['metadata']['payment_number'] = ''
+                if 'order_number' in provider_response_body['metadata']:
+                    order_uuid = provider_response_body['metadata']['order_number']
```
HammadAhmadWaqas, line 108: why we are getting order_uuid here from `provider_response_body`? `order_uuid` is already available here that was extracted from Stripe event metadata.

commerce_coordinator/apps/titan/tasks.py
```diff
@@ -102,8 +101,22 @@ def payment_processed_save_task(
...
+            titan_api_client.create_payment(
+                order_uuid=order_uuid,
+                reference_number=reference_number,
+                payment_method_name=PaymentMethod.STRIPE.value,
+                provider_response_body=provider_response_body,
```
HammadAhmadWaqas, line 113: This `provider_response_body` is Stripe event with failed payment information. We are already saving this information to Titan system with our failed payment (l[ine 96](https://github.com/edx/commerce-coordinator/pull/107/files#diff-bf55193548ab769a028ca42527203ecd61a500970277a1881d13390589e12a54L97))

In the original follow, We use the payload that we get from Stripe while creating payment intent. As we don't have that here, using `provider_response_body=None` would make more sense.

@pshiu what would you say?

commerce_coordinator/apps/titan/tasks.py
```diff
@@ -102,8 +101,22 @@ def payment_processed_save_task(
             payment_state_paid_cache_key = get_paid_payment_state_cache_key(payment_number)
             TieredCache.set_all_tiers(payment_state_paid_cache_key, payment, settings.DEFAULT_TIMEOUT)
         elif payment_state == PaymentState.FAILED.value:
+            if 'metadata' in provider_response_body:
+                if 'payment_number' in provider_response_body['metadata']:
+                    provider_response_body['metadata']['payment_number'] = ''
```
HammadAhmadWaqas, line 106: What are we doing here?

commerce_coordinator/apps/titan/tasks.py
HammadAhmadWaqas: or maybe we can use `StripeApiClient.retrieve_payment_intent` to get provide_response_body.

commerce_coordinator/apps/titan/tasks.py
wdrussell2015: removed overlooked that it was already coming from the stripe event

commerce_coordinator/apps/titan/exceptions.py
```diff
@@ -21,6 +21,12 @@ class ProcessingAlreadyRequested(APIException):
     default_code = 'already_processing'
 
 
+class StaleOrder(APIException):
```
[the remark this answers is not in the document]
wdrussell2015: misread the api spec not needed, removed

commerce_coordinator/apps/titan/tasks.py
wdrussell2015: changed to retrieve call since provider_response_body is a required field

commerce_coordinator/apps/titan/tasks.py
wdrussell2015: Removing the old payment number from the stripe response metadata so its not saved in the new payment, I guess it isnt really necessary since well update that information when the user clicks Pay again and isnt used in between those events

HammadAhmadWaqas (approved):

PR #111, opened 2023-08-15
chore: Fixing use of `Enum` as Strings (have to call `.value`)

grmartin:
- Text fixes too.


commerce_coordinator/apps/lms/tests/test_views.py
```diff
@@ -317,11 +317,13 @@ def test_create_order(
                 expected_error_or_response
             )
 
+            query_params = parse_qs(urlparse(response.headers['Location']).query)
```
grmartin, line 320: Note: parse_qs returns values as arrays, just incase you use the key[]=1&key[]=2 syntax.

PR #112, opened 2023-08-16
build: Remove pycrypto system lib deletion hack

timmc-edx:
It's not clear what this was for, but it's probably not needed and doesn't seem like it would work anyhow. (And it's making `make requirements` fail.)

Part of https://github.com/edx/edx-arch-experiments/issues/391


robrap:
> The tests are failing because ci.in still lists codecov, which was yanked from PyPI. There are also pip/pip-tools version conflicts.

I'm not clear on what this means, but it sounds like you are saying that this PR won't make it any worse?

pshiu (approved):

pshiu:
These changes also in branch `2u/project-theseus` in https://github.com/edx/commerce-coordinator/commit/47c591ad5952b6a0b710d83e60f5c2ca3cbff8d3.

PR #128, opened 2023-11-15
feat: Checks for redirecting to CT checkout

JadeyOlivier:
[WIP]
**Description:** Conditionally routing learners to the CT checkout page based on waffle toggle status. 

**JIRA:** [SONIC-100](https://2u-internal.atlassian.net/browse/SONIC-100)

**Dependencies:** dependencies on other outstanding PRs, issues, etc.

### DIFF
--- a/commerce_coordinator/apps/rollout/pipeline.py
+++ b/commerce_coordinator/apps/rollout/pipeline.py
@@ -1,0 +1,44 @@
+import logging
+
+from openedx_filters import PipelineStep
+from openedx_filters.exceptions import OpenEdxFilterException
+
+from commerce_coordinator.apps.commercetools_frontend.constants import COMMERCETOOLS_FRONTEND
+from commerce_coordinator.apps.frontend_app_payment.constants import FRONTEND_APP_PAYMENT_CHECKOUT
+from commerce_coordinator.apps.rollout.waffle import is_redirect_to_commercetools_enabled_for_user
+
+logger = logging.getLogger(__name__)
+
+
+class GetActiveOrderManagementSystem(PipelineStep):
+    """
+    Determines which order management system to redirect to, based on the existence of course_run_id
+    and waffle flag's value.
+    """
+
+    def run_filter(self, request, active_order_management_system):
+        """
+        Execute a filter with the signature specified.
+        Arguments:
+            request: request object passed through from the lms filter
+            active_order_management_system: result from pipeline steps to determine if
+                            redirect_to_commercetools_checkout flag is on and course_run_key
+                            query param exists & dictact which checkout to redirect to
+        """
+
+        sku = request.query_params.get('sku', '').strip()
+        course_run = request.query_params.get('course_run_key', '').strip()
+
+        if is_redirect_to_commercetools_enabled_for_user(request) and course_run:
+            active_order_management_system = COMMERCETOOLS_FRONTEND
+        elif sku:
+            active_order_management_system = FRONTEND_APP_PAYMENT_CHECKOUT
+        else:
+            logger.exception(f'An error occurred while determining the active order management system.'
+                             f'No waffle flag, course_run_key or sku value found')
+            raise OpenEdxFilterException(f'Neither course_run_key and waffle flag value nor sku found.'
+                                         f'Unable to determine active order management system.')
+
+        return {
+            'active_order_management_system': active_order_management_system
+        }
--- a/commerce_coordinator/apps/rollout/waffle.py
+++ b/commerce_coordinator/apps/rollout/waffle.py
@@ -1,0 +1,16 @@
+"""
+Configuration settings via waffle flags for redirecting to CommerceTools.
+"""
+
+import waffle  # pylint: disable=invalid-django-waffle-import
+
+WAFFLE_FLAG_NAMESPACE = "transition_to_commercetools"
+
+REDIRECT_TO_COMMERCETOOLS_CHECKOUT = f'{WAFFLE_FLAG_NAMESPACE}.redirect_to_commercetools_checkout'
+
+
+def is_redirect_to_commercetools_enabled_for_user(request):
+    """
+    Check if REDIRECT_TO_COMMERCETOOLS_CHECKOUT flag is enabled.
+    """
+    return waffle.flag_is_active(request, REDIRECT_TO_COMMERCETOOLS_CHECKOUT)
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -3,7 +3,6 @@ API clients for commercetools app.
 """
 
 import logging
-import typing
 from typing import Generic, List, Optional, TypeVar
 
 import requests
--- a/commerce_coordinator/apps/frontend_app_ecommerce/tests/test_filters.py
+++ b/commerce_coordinator/apps/frontend_app_ecommerce/tests/test_filters.py
@@ -27,7 +27,6 @@ class TestFrontendAppEcommerceFilters(TestCase):
     """
 
     # Disable unused-argument due to global @patch
-    # pylint: disable=unused-argument
 
     def test_order_history_requested_filter(self, _mock_ctorders, _mock_ecommerce_client):
         """
--- a/commerce_coordinator/apps/frontend_app_ecommerce/tests/test_views.py
+++ b/commerce_coordinator/apps/frontend_app_ecommerce/tests/test_views.py
@@ -30,7 +30,6 @@ class OrdersViewTests(TestCase):
     Tests for order views.
     """
     # Disable unused-argument due to global @patch
-    # pylint: disable=unused-argument
 
     # Use Django Rest Framework client for self.client
     client_class = APIClient


commerce_coordinator/apps/rollout/constants.py
```diff
@@ -0,0 +1,9 @@
+# .. toggle_name: redirect_to_commercetools_checkout
+# .. toggle_type: waffle_flag
+# .. toggle_default: False
+# .. toggle_description: Determines if user will be redirected to commerce tools or legacy
+# .. toggle_creation_date: 2023-11-15
+# .. toggle_tickets: SONIC-100
+
+# TODO: Create flag in stage/prod
+REDIRECT_TO_COMMERCETOOLS_CHECKOUT = 'redirect_to_commercetools_checkout'
```
grmartin, line 9: Hiya! So we keep these listed in Confluence and have a format for the individual flag. I will DM you the page.

shafqatfarhan, line 9: Can we please move waffle flags to a separate file instead of constants? Ref: https://github.com/edx/subscriptions/blob/main/subscriptions/apps/subscriptions/waffle.py

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+            active_om_system: result from pipeline steps to determine if user has
+                            enterprise role or not & dictact which checkout to redirect to
+        """
+
+        sku = request.query_params.get('sku', '').strip()
+        sku_exists = isinstance(sku, str) and sku  # True if sku not None and is string
+        course_run = request.query_params.get('course_run_key', '').strip()
+        course_run_exists = isinstance(course_run, str) and course_run  # True if course_run_key not None and is string
+
+        if waffle.flag_is_active(request, REDIRECT_TO_COMMERCETOOLS_CHECKOUT) and sku_exists and course_run_exists:
```
shafqatfarhan, line 32: It would be better if we move this logic to a utility function created in waffle.py and simply call the function here or anywhere else required.
```suggestion
        if is_redirect_to_commercetools_enabled_for_user(request) and sku_exists and course_run_exists:
```

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        Execute a filter with the signature specified.
+        Arguments:
+            request: request object passed through from the lms filter
+            active_om_system: result from pipeline steps to determine if user has
+                            enterprise role or not & dictact which checkout to redirect to
+        """
+
+        sku = request.query_params.get('sku', '').strip()
+        sku_exists = isinstance(sku, str) and sku  # True if sku not None and is string
```
shafqatfarhan, line 28: `isinstance(sku, str)` This function would still be True for `''` empty string. We could simply do: 
```suggestion
        sku_exists = True if sku else False
```

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+from commerce_coordinator.apps.commercetools_frontend.constants import COMMERCETOOLS_FRONTEND
+from commerce_coordinator.apps.frontend_app_payment.constants import FRONTEND_APP_PAYMENT_CHECKOUT
+from commerce_coordinator.apps.rollout.constants import REDIRECT_TO_COMMERCETOOLS_CHECKOUT
+
+logger = logging.getLogger(__name__)
+
+
+class GetActiveOrderManagementSystem(PipelineStep):
+    """
+    Determines which order management system to redirect to, based on enterprise role.
```
shafqatfarhan, line 15: Can we update the description to reflect that order management is determined based on the existence of course_run_id and waffle flag's value?

commerce_coordinator/apps/rollout/pipeline.py
shafqatfarhan, line 32: We don't rely on `sku` to decide on CT as an active system.
```suggestion
        if is_redirect_to_commercetools_enabled_for_user(request) and course_run:
```

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+
+        sku = request.query_params.get('sku', '').strip()
+        sku_exists = isinstance(sku, str) and sku  # True if sku not None and is string
+        course_run = request.query_params.get('course_run_key', '').strip()
+        course_run_exists = isinstance(course_run, str) and course_run  # True if course_run_key not None and is string
+
+        if waffle.flag_is_active(request, REDIRECT_TO_COMMERCETOOLS_CHECKOUT) and sku_exists and course_run_exists:
+            active_om_system = COMMERCETOOLS_FRONTEND
+        else:
+            active_om_system = FRONTEND_APP_PAYMENT_CHECKOUT
```
shafqatfarhan, line 35:
```suggestion
        elif sku:
            active_om_system = FRONTEND_APP_PAYMENT_CHECKOUT
        else:
            <log and raise error>
```

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return {
+            'request': request + {'active_om_system': active_om_system}
```
shafqatfarhan, line 38: nit: active_order_management_system

shafqatfarhan (changes requested):

commerce_coordinator/apps/rollout/pipeline.py
JadeyOlivier: Is there no need to check if the course_run is not none and a string? Are we simply checking the availability of the `course_run_key` query param?

commerce_coordinator/apps/rollout/pipeline.py
JadeyOlivier: If we are only checking `sku` here, is there a need for `sku_exists`?

commerce_coordinator/apps/rollout/constants.py
shafqatfarhan: lets add it in `app/rollout`

commerce_coordinator/apps/rollout/pipeline.py
shafqatfarhan: yes.

JadeyOlivier: [Update] We've decided we don't need to explicitly check the instance of the query parameters to be a string as the line above does covers this requirement, so we can simply check the value of `sku` and `course_run` .

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,42 @@
+import logging
```
JadeyOlivier, line 1: Please note that the tests are failing as the constants files reside in this [PR](https://github.com/edx/commerce-coordinator/pull/129).

```
from commerce_coordinator.apps.commercetools_frontend.constants import COMMERCETOOLS_FRONTEND
E   ModuleNotFoundError: No module named 'commerce_coordinator.apps.commercetools_frontend'
```

I temporarily created these constants files and ran the tests to ensure they were working on my local and all of them passed, so the changes in the above PR are added to mine there should be no issues :slightly_smiling_face:

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,42 @@
...
+            active_order_management_system = COMMERCETOOLS_FRONTEND
+        elif sku:
+            active_order_management_system = FRONTEND_APP_PAYMENT_CHECKOUT
+        else:
+            logger.exception(f'An error occurred while determining the active order management system.'
+                             f'No waffle flag, course_run_key or sku value found')
+            raise Exception('Neither course_run_key and waffle flag value nor sku found. Unable to determine active order management system.')
+
+        return {
+            'request': dict(request.GET, active_order_management_system=active_order_management_system),
```
shafqatfarhan, line 41: Why are we returning `request.GET` dict? I believe we should return the desired key/value pair only as the request context is available in the Django view.

commerce_coordinator/apps/rollout/waffle.py
```diff
@@ -0,0 +1,16 @@
...
+import waffle  # pylint: disable=invalid-django-waffle-import
+
+WAFFLE_FLAG_NAMESPACE = "transition_to_commercetools"
+
+REDIRECT_TO_COMMERCETOOLS_CHECKOUT = f'{WAFFLE_FLAG_NAMESPACE}.redirect_to_commercetools_checkout'
+
+
+def is_redirect_to_commercetools_enabled_for_user(request):
+    """
+    Check if OVERRIDE_SUBSCRIPTION_TRIAL_END_DATE flag is enabled.
```
shafqatfarhan, line 14:
```suggestion
    Check if REDIRECT_TO_COMMERCETOOLS_CHECKOUT flag is enabled.
```

shafqatfarhan (commented):
Left a few suggestions.

commerce_coordinator/apps/rollout/waffle.py
JadeyOlivier: Sorry, small overlook, will rectify.

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -0,0 +1,43 @@
...
+        elif sku:
+            active_order_management_system = FRONTEND_APP_PAYMENT_CHECKOUT
+        else:
+            logger.exception(f'An error occurred while determining the active order management system.'
+                             f'No waffle flag, course_run_key or sku value found')
+            raise OpenEdxFilterException('Neither course_run_key and waffle flag value nor sku found. Unable to determine active order management system.')
+
+        return {
+            'request': dict(request, active_order_management_system=active_order_management_system),
+        }
```
shafqatfarhan, line 43:
```suggestion
        return {
            'active_order_management_system': active_order_management_system
        }
```

shafqatfarhan (approved):
LGTM 👍

pshiu (approved):
Fantastic PR, thanks @JadeyOlivier!

PR #129, opened 2023-11-16
feat: Created endpoint to route learners to correct checkout page

MushtaqRossier:
**Description:** [WIP] Created endpoint to route learners to correct checkout page

**JIRA:** [SONIC-122](https://2u-internal.atlassian.net/browse/SONIC-122)

### DIFF
--- a/commerce_coordinator/apps/commercetools_frontend/pipeline.py
+++ b/commerce_coordinator/apps/commercetools_frontend/pipeline.py
@@ -1,0 +1,23 @@
+"""
+Commercetools filter pipeline implementation
+"""
+
+from django.conf import settings
+from openedx_filters import PipelineStep
+
+from commerce_coordinator.apps.commercetools_frontend.constants import COMMERCETOOLS_FRONTEND
+from commerce_coordinator.apps.core.constants import PipelineCommand
+
+
+class GetCommercetoolsRedirectUrl(PipelineStep):
+    """
+    Generates Commercetools redirect url
+    """
+
+    def run_filter(self, active_order_management_system, request):
+        if active_order_management_system == COMMERCETOOLS_FRONTEND:
+            return {
+                'redirect_url': settings.COMMERCETOOLS_FRONTEND_URL
+            }
+        else:
+            return PipelineCommand.CONTINUE.value
--- a/commerce_coordinator/apps/frontend_app_payment/pipeline.py
+++ b/commerce_coordinator/apps/frontend_app_payment/pipeline.py
@@ -1,0 +1,23 @@
+"""
+Legacy filter pipeline implementation
+"""
+
+from django.conf import settings
+from openedx_filters import PipelineStep
+
+from commerce_coordinator.apps.core.constants import PipelineCommand
+from commerce_coordinator.apps.frontend_app_payment.constants import FRONTEND_APP_PAYMENT_CHECKOUT
+
+
+class GetPaymentMFERedirectUrl(PipelineStep):
+    """
+    Generates legacy redirect url
+    """
+    def run_filter(self, active_order_management_system, request, redirect_url=None):
+        # pylint: disable=arguments-differ, unused-argument
+        if active_order_management_system == FRONTEND_APP_PAYMENT_CHECKOUT:
+            return {
+                "redirect_url": settings.FRONTEND_APP_PAYMENT_URL
+            }
+        else:
+            return PipelineCommand.CONTINUE.value
--- a/commerce_coordinator/apps/lms/filters.py
+++ b/commerce_coordinator/apps/lms/filters.py
@@ -5,21 +5,21 @@ Filters for LMS
 from openedx_filters.tooling import OpenEdxPublicFilter
 
 
-class OrderCreateRequested(OpenEdxPublicFilter):
+class PaymentPageRedirectRequested(OpenEdxPublicFilter):
     """
-    Filter to gather order data from the defined PipelineStep(s)
+    Filter to gather payment page redirect urls.
     """
     # See pipeline step configuration OPEN_EDX_FILTERS_CONFIG dict in `settings/base.py`
-    filter_type = "org.edx.coordinator.lms.order.create.requested.v1"
+    filter_type = "org.edx.coordinator.lms.payment.page.redirect.requested.v1"
 
     @classmethod
-    def run_filter(cls, params, order_data=None):
+    def run_filter(cls, request):
         """
-        Call the PipelineStep(s) defined for this filter, to gather orders and return together
+        Execute the filter pipeline with the desired signature.
         Arguments:
-            params: arguments passed through from the original order history url querystring
-            order_data: any preliminary orders we want to append to when running the pipeline
+            request: django.http.HttpRequest object passed through from lms views.py
         """
 
-        # super op
-        return super().run_pipeline(params=params, order_data=order_data)
+        redirect_url = super().run_pipeline(request=request)
+
+        return redirect_url
--- a/commerce_coordinator/apps/lms/tests/test_views.py
+++ b/commerce_coordinator/apps/lms/tests/test_views.py
@@ -1,75 +1,28 @@
 """
 Tests for the LMS (edx-platform) views.
 """
-import uuid
-from urllib.parse import parse_qs, urlparse
 
 import ddt
-import django.conf
+from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.test import override_settings
 from django.urls import reverse
 from mock import patch
-from openedx_filters import PipelineStep
 from rest_framework import status
 from rest_framework.test import APITestCase
-
-from commerce_coordinator.apps.core.constants import QueryParamPrefixes, WaffleFlagNames
-from commerce_coordinator.apps.core.tests.utils import name_test, uuid4_str
-from commerce_coordinator.apps.lms.filters import OrderCreateRequested
 
 User = get_user_model()
 
 
-class TestOrderCreateRequestedFilterPipelineThatExplodes(PipelineStep):
-    """
-    An example exploding Pipeline, to test filter failures return expected codes when if something is uncaught
-    """
-
-    targets = 'org.edx.coordinator.lms.order.create.requested.v1'
-
-    @classmethod
-    def get_fqtn(cls):
-        """ Return the fully qualified type name to this class """
-        return f'{cls.__module__}.{cls.__qualname__}'
-
-    def run_filter(self, params, order_data):  # pylint: disable=arguments-differ
-        """ Implemented by the default pipeline, this intentionally explodes on us. Has a bomb emoji too. """
-        raise Exception('\U0001f4a3 (this is intentional)')
-
-
-class TestOrderCreateRequestedFilterPipelineRecordsInputData(PipelineStep):
-    """
-    An example no-op Pipeline, to test filter success and validate result
-    """
-
-    LAST_DATA = None
-    targets = 'org.edx.coordinator.lms.order.create.requested.v1'
-
-    @classmethod
-    def get_fqtn(cls):
-        """ Return the fully qualified type name to this class """
-        return f'{cls.__module__}.{cls.__qualname__}'
-
-    def run_filter(self, params, order_data):  # pylint: disable=arguments-differ
-        """ Implemented by the default pipeline, this intentionally doesnt do much, but allows introspection. """
-        TestOrderCreateRequestedFilterPipelineRecordsInputData.LAST_DATA = {
-            'params': dict(params),
-            'order_data': order_data
-        }
-        return TestOrderCreateRequestedFilterPipelineRecordsInputData.LAST_DATA
-
-
 @ddt.ddt
-class OrderCreateViewTests(APITestCase):
+class PaymentPageRedirectViewTests(APITestCase):
     """
-    Tests for order create view.
+    Tests for payment page redirect view.
     """
     # Define test user properties
     test_user_username = 'test'
     test_user_email = 'test@example.com'
     test_user_password = 'secret'
-    url = reverse('lms:create_order')
+    url = reverse('lms:payment_page_redirect')
 
     def setUp(self):
         """Create test user before test starts."""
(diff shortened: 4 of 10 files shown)


commerce_coordinator/settings/base.py
```diff
@@ -369,6 +369,13 @@ def root(*path_fragments):
             'commerce_coordinator.apps.titan.pipeline.CreateTitanOrder',
         ]
     },
+    "org.edx.coordinator.lms.payment.page.redirect.requested.v1": {
+        "fail_silently": False,  # TODO: Coordinator filters should NEVER be allowed to fail silently
+        "pipeline": [
+            'commerce_coordinator.apps.commercetools_frontend.pipeline.GetCommercetoolsRedirectUrl',
+            'commerce_coordinator.apps.lms.pipeline.GetLegacyRedirectUrl'
```
shafqatfarhan, line 376:
```suggestion
            'commerce_coordinator.apps.frontend_app_payment.pipeline.GetLegacyRedirectUrl'
```

shafqatfarhan (changes requested):

commerce_coordinator/apps/lms/urls.py
```diff
@@ -4,9 +4,9 @@
 
 from django.urls import path
 
-from commerce_coordinator.apps.lms.views import OrderCreateView
+from commerce_coordinator.apps.lms.views import PaymentPageRedirectView
 
 app_name = 'lms'
 urlpatterns = [
-    path('order/', OrderCreateView.as_view(), name='create_order'),
+    path('redirect/', PaymentPageRedirectView.as_view(), name='payment_page_redirect'),
```
aht007, line 11:
```suggestion
    path('payment_page_redirect', PaymentPageRedirectView.as_view(), name='payment_page_redirect'),
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -4,76 +4,53 @@
...
 from rest_framework.status import HTTP_303_SEE_OTHER
 from rest_framework.throttling import UserRateThrottle
 from rest_framework.views import APIView
 
-from ..core.constants import HttpHeadersNames, MediaTypes, QueryParamPrefixes, WaffleFlagNames
-from .filters import OrderCreateRequested
-from .serializers import OrderCreatedSignalInputSerializer
+from commerce_coordinator.apps.lms.filters import PaymentPageRedirectRequested
+
+from ..core.constants import HttpHeadersNames, MediaTypes, QueryParamPrefixes
```
aht007, line 15:
```suggestion
from commerce_coordinator.apps.core.constants import HttpHeadersNames, MediaTypes, QueryParamPrefixes
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -4,76 +4,53 @@
...
 
-        Query Parameters:
-            sku: Array. An edx.org stock keeping units (SKUs) that the user would like to purchase.
-            coupon_code: (Optional) A coupon code to initially apply to the order.
-
         Returns:
             an HTTP Response, in the form of an error or as an HTTP Redirect.
 
-            - HTTP Redirect (303): Redirection to Payment MFE upon Success
+            - HTTP Redirect (303): Redirection to Commercetools checkout upon Success
```
aht007, line 35: Based on conditional, this will redirect to both CT and legacy checkout. We should correct this information in the docstring.

commerce_coordinator/apps/lms/views.py
```diff
@@ -4,76 +4,53 @@
...
-                logger.exception(f"Something went wrong! Exception raised in {self.get.__name__} with error {repr(e)}")
-                return HttpResponseServerError()
-
-        # Log for Diagnostics
-        logger.debug(f'{self.get.__qualname__} we didnt redirect..')  # pragma: no cover
+        try:
+            return self._redirect_response_payment(request)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.exception(f"Something went wrong! Exception raised in {self.get.__name__} with error {repr(e)}")
+            return HttpResponseServerError()
```
aht007, line 49: We should instead return: HttpResponseBadRequest('Something went wrong.')

commerce_coordinator/apps/lms/views.py
```diff
@@ -82,10 +59,12 @@ def _redirect_response_payment(self, request):
         """
 
         get_items = list(self.request.GET.items())
-        get_items.append((f"{QueryParamPrefixes.WAFFLE_FLAG.value}{WaffleFlagNames.COORDINATOR_ENABLED.value}", '1'))
+        get_items.append((f"{QueryParamPrefixes.WAFFLE_FLAG.value}", '1'))
```
aht007, line 62: Is there any need to send this waffle flag value in query params?

commerce_coordinator/apps/lms/views.py
```diff
@@ -95,20 +74,18 @@ def _redirect_response_payment(self, request):
         return redirect
 
     @staticmethod
-    def _add_utm_and_waffle_flags_params_to_url(url, params):
+    def _add_utm_params_to_url(url, params):
```
aht007, line 77: This function is just adding the utm params which isn't what we intend to do.
We should extract the query params that were sent to coordinator with our request and append those query params to our redirect url. You can make a new utility for this purpose(`_add_query_params_to_redirect_url`).

commerce_coordinator/apps/lms/filters.py
```diff
@@ -4,22 +4,19 @@
...
-        Call the PipelineStep(s) defined for this filter, to gather orders and return together
-        Arguments:
-            params: arguments passed through from the original order history url querystring
-            order_data: any preliminary orders we want to append to when running the pipeline
+        Execute the filter pipeline with the desired signature.
         """
 
-        # super op
-        return super().run_pipeline(params=params, order_data=order_data)
+        rollout_data = rollout.pipeline.GetActiveOrderManagementSystem.run_pipeline(params)
```
aht007, line 20: Can't we make this a step in the pipeline as well? Why do we need to execute it explicitly? Can you verify this

aht007 (changes requested):

commerce_coordinator/apps/lms/filters.py
MushtaqRossier: This is the pipeline created in @JadeyOlivier ticket. Based on the result of her pipeline, that determines which redirect url gets build

commerce_coordinator/apps/lms/views.py
MushtaqRossier: I'm extracting the query params from the request object in the function above and passing it in as the argument 'params' in this function.

MushtaqRossier: The argument 'params' is a list of all query params extracted from the request url

commerce_coordinator/apps/lms/views.py
```diff
@@ -4,76 +4,53 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-class OrderCreateView(APIView):
-    """Accept incoming request for creating a basket/order for a user."""
+class PaymentPageRedirectView(APIView):
+    """Accept incoming request for routing users to checkout view."""
     permission_classes = (LoginRedirectIfUnauthenticated,)
     throttle_classes = (UserRateThrottle,)
```
aht007, line 23: Can you double check if the throttle rate settings are the same as legacy ecommerce endpoint?
Specifically I didn't find any default throttle rate settings in CC.
For reference: https://github.com/openedx/ecommerce/blob/master/ecommerce/settings/base.py#L527

commerce_coordinator/apps/lms/views.py
```diff
@@ -82,10 +59,12 @@ def _redirect_response_payment(self, request):
         """
 
         get_items = list(self.request.GET.items())
-        get_items.append((f"{QueryParamPrefixes.WAFFLE_FLAG.value}{WaffleFlagNames.COORDINATOR_ENABLED.value}", '1'))
+        get_items.append((f"{QueryParamPrefixes.GOOGLE_ANALYTICS.value}", '1'))
```
aht007, line 62: I still don't get why do we need google analytics query param in our new CC redirect URL as I previously mentioned. Nor I could find any such information in the ticket. Can you please mention the need for it?

commerce_coordinator/apps/lms/views.py
```diff
@@ -95,24 +74,22 @@ def _redirect_response_payment(self, request):
...
             params (list): Query Params from Req as an encoded list
         Returns:
             url (str): A URL asa Python String
         """
 
-        # utm_params is [(u'utm_content', u'course-v1:IDBx IDB20.1x 1T2017'),... + waffles
-        utm_params = [item for item in params
-                      if QueryParamPrefixes.GOOGLE_ANALYTICS.value or QueryParamPrefixes.WAFFLE_FLAG.value in item[0]]
+        # utm_params is [(u'utm_content', u'course-v1:IDBx IDB20.1x 1T2017'),...
+        query_params = [item for item in params
```
aht007, line 87: My previous suggestion was to add a new utility and keep this one as it is since we were not going to use it. My expectation was that in our new utility we would just append the query params that we received from the track selection page as it is. We don't need to do any conditional check as far as I understand nor do we need to any additional params(e.g. google analytics etc). We just send what we receive as it is(which would just be the sku and course_run_key IMO).
That is my understanding.

aht007 (changes requested):
I have added some followup comments. I still see some unresolved comments/suggestions from the previous iteration. Please resolve those as well. Also there are some conflicts in the PR that need to addressed.

commerce_coordinator/apps/lms/views.py
MushtaqRossier: Can confirm that it's the same

aht007: I didn't find any default throttle rate setting in the settings files. Can you point me to that?

commerce_coordinator/apps/lms/filters.py
```diff
@@ -4,22 +4,19 @@
...
-    Filter to gather order data from the defined PipelineStep(s)
+    Filter to gather payment page redirect urls.
     """
     # See pipeline step configuration OPEN_EDX_FILTERS_CONFIG dict in `settings/base.py`
-    filter_type = "org.edx.coordinator.lms.order.create.requested.v1"
+    filter_type = "org.edx.coordinator.lms.payment.page.redirect.requested.v1"
 
     @classmethod
-    def run_filter(cls, params, order_data=None):
+    def run_filter(cls, params):
```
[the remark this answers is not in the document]
aht007: https://github.com/edx/commerce-coordinator/pull/129/files#diff-29065c51c7310dc90c483b5e477daa29aa7a9efcbba0570397035abc88ae6148R64
From here we are sending request object, but we have named it params here

commerce_coordinator/apps/lms/views.py
MushtaqRossier: https://github.com/edx/commerce-coordinator/blob/2u/project-replatforming/commerce_coordinator/settings/base.py#L283

commerce_coordinator/apps/lms/views.py
MushtaqRossier: I had a discussion with @shafqatfarhan and we agreed that rather than adding a new utility function, we can modify this current one to add all query params to the redirect url. Reason being, all the previous code in this file is no longer needed hence modifying this endpoint to satisfy our needs.

commerce_coordinator/settings/base.py
```diff
@@ -348,7 +348,11 @@ def root(*path_fragments):
 # Needed to link to the payment micro-frontend
 PAYMENT_MICROFRONTEND_URL = 'replace-me'
 
-# Coordinator filters should NEVER be allowed to fail silently
+# Checkout view urls
+COMMERCETOOLS_FRONTEND_URL = 'https://checkout.edx.org'
```
aht007, line 352: We should add values for local development here. Prod and stage config should be overridden from here: [edx-internal link](https://github.com/edx/edx-internal/tree/4b294b211801c57979c820795eea4e555b2597fd/argocd/applications/commerce-coordinator)

aht007 (approved):

commerce_coordinator/apps/frontend_app_payment/pipeline.py
```diff
@@ -0,0 +1,24 @@
...
+from commerce_coordinator.apps.frontend_app_payment.constants import FRONTEND_APP_PAYMENT_CHECKOUT
+
+
+class GetLegacyRedirectUrl(PipelineStep):
+    """
+    Generates legacy redirect url
+    """
+    def run_filter(self, active_order_management_system ,redirect_url):
+
+        if redirect_url is not None or active_order_management_system['active_order_management_system'] != FRONTEND_APP_PAYMENT_CHECKOUT:
```
aht007, line 17: I suppose redirect_url is a dictionary returned from first pipeline step. Isn't that the case?
```suggestion
        if redirect_url.get(redirect_url, None) is not None or active_order_management_system['active_order_management_system'] != FRONTEND_APP_PAYMENT_CHECKOUT:
```

commerce_coordinator/settings/base.py
```diff
@@ -348,7 +348,11 @@ def root(*path_fragments):
...
+FRONTEND_APP_PAYMENT_URL = 'https://payment.edx.org'
```
aht007, line 353:
```suggestion
FRONTEND_APP_PAYMENT_URL = 'https://payment_page_url'
```

commerce_coordinator/settings/base.py
```diff
@@ -348,7 +348,11 @@ def root(*path_fragments):
 # Needed to link to the payment micro-frontend
 PAYMENT_MICROFRONTEND_URL = 'replace-me'
 
-# Coordinator filters should NEVER be allowed to fail silently
+# Checkout view urls
+COMMERCETOOLS_FRONTEND_URL = 'https://commerce_tools_frontend_url'
+FRONTEND_APP_PAYMENT_URL = 'https://payment_page_url'
+
+# Filters PoC
```
shafqatfarhan, line 355: Can you check if this comment is required?

commerce_coordinator/apps/frontend_app_payment/pipeline.py
```diff
@@ -0,0 +1,25 @@
...
+from openedx_filters import PipelineStep
+
+from commerce_coordinator.apps.frontend_app_payment.constants import FRONTEND_APP_PAYMENT_CHECKOUT
+
+
+class GetLegacyRedirectUrl(PipelineStep):
+    """
+    Generates legacy redirect url
+    """
+    def run_filter(self, active_order_management_system, redirect_url):  # pylint: disable=arguments-differ
```
shafqatfarhan, line 15: Can you confirm if the return property from 1st pipeline step (i.e. `redirect_url`) is propagated here to this (2nd) pipeline step as a method argument?

shafqatfarhan (approved):

commerce_coordinator/apps/frontend_app_payment/pipeline.py
MushtaqRossier: I can confirm that it is.

commerce_coordinator/settings/base.py
MushtaqRossier: I'm not too sure. @grmartin could you comment on this?

pshiu (changes requested):

commerce_coordinator/apps/commercetools_frontend/pipeline.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class GetCommercetoolsRedirectUrl(PipelineStep):
+    """
+    Generates Commercetools redirect url
+    """
+
+    def run_filter(self, active_order_management_system):
+
+        redirect_url = None
```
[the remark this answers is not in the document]
pshiu: This still appears to be an issue - you may want to copy your logic from frontend_app_payment/pipeline.py here.

commerce_coordinator/apps/lms/filters.py
```diff
@@ -5,21 +5,19 @@
...
 
     @classmethod
-    def run_filter(cls, params, order_data=None):
+    def run_filter(cls, request):
         """
-        Call the PipelineStep(s) defined for this filter, to gather orders and return together
-        Arguments:
-            params: arguments passed through from the original order history url querystring
-            order_data: any preliminary orders we want to append to when running the pipeline
+        Execute the filter pipeline with the desired signature.
```
pshiu, line 18: Can we docstring what `request` is?

commerce_coordinator/apps/lms/views.py
```diff
@@ -4,76 +4,52 @@
...
 
-        Query Parameters:
-            sku: Array. An edx.org stock keeping units (SKUs) that the user would like to purchase.
-            coupon_code: (Optional) A coupon code to initially apply to the order.
-
         Returns:
             an HTTP Response, in the form of an error or as an HTTP Redirect.
 
-            - HTTP Redirect (303): Redirection to Payment MFE upon Success
+            - HTTP Redirect (303): Redirection to Commercetools or Legacy checkout upon Success
```
pshiu, line 34: Nit: Make more generic so it's independent of pipeline configuration:

```suggestion
            - HTTP Redirect (303): Redirection to correct checkout page upon Success
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -4,76 +4,52 @@
...
             - HTTP File or Server Error (400-599) see Errors section for more information.
 
         Errors:
-            - 400: if required params are missing or not in supported format.
```
pshiu, line 46: A 400 is still possible, right? Via `HttpResponseBadRequest` [[Django docs](https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpResponseBadRequest)] in line 48.

commerce_coordinator/apps/lms/views.py
```diff
@@ -95,24 +72,18 @@ def _redirect_response_payment(self, request):
...
-        # utm_params is utm_content=course-v1:IDBx+IDB20.1x+1T2017&...
-        # (course-keys do not have url encoding)
-        utm_params = unquote(utm_params)
-        url = url + '?' + utm_params if utm_params else url
+        query_params = list(params)
+        query_params = urlencode(query_params, True)
+        query_params = unquote(query_params)
+        url = url + '?' + query_params if query_params else url
+
         return url
```
pshiu, line 89: Would just like to get sign off from @grmartin that this code & related tests are no longer needed - (sorry if I missed an earlier conversation, but I thought we were still going to need Analytics for our release).

commerce_coordinator/apps/lms/tests/test_views.py
pshiu, line 1: There seem to be a lot of test deletions in this file - good to clean up, but please check that none are applicable/transferrable to our current use case. This will save us time later in improving code coverage.

pshiu (commented):
Super appreciate the endurance on this one. A couple more:

commerce_coordinator/apps/lms/views.py
MushtaqRossier: Correct yes. Will add it back it.

commerce_coordinator/apps/lms/views.py
MushtaqRossier: Had chats with @shafqatfarhan and @aht007 that the purpose of this method is to just extract query params from request url and append to it redirect url. The conversations should fall under the resolved chats in this pr.

commerce_coordinator/apps/lms/tests/test_views.py
MushtaqRossier: There were a lot of unneeded tests for order creation and pipeline validations. Had discussions with @shafqatfarhan  that the unit test for this work should only test whether the redirect was successful or not.

pshiu: Not sure that [these tests](https://github.com/edx/commerce-coordinator/pull/129/files#diff-d887746bdaeac8ad9988169c12b3c8c6f59754d7c6977c664980a1d53a58af97L90-L106) were unneeded? [Coverage also dropped](https://app.codecov.io/gh/edx/commerce-coordinator/pull/129/blob/commerce_coordinator/apps/lms/views.py#L47) due to one of the deleted tests.

grmartin (approved):

commerce_coordinator/apps/frontend_app_payment/pipeline.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from django.conf import settings
+from openedx_filters import PipelineStep
+
+from commerce_coordinator.apps.core.constants import PipelineCommand
+from commerce_coordinator.apps.frontend_app_payment.constants import FRONTEND_APP_PAYMENT_CHECKOUT
+
+
+class GetLegacyRedirectUrl(PipelineStep):
```
pshiu, line 12: Can we call this `GetPaymentMFERedirect`, [as spec'ed](https://github.com/edx/commerce-coordinator-private/pull/83)? "Legacy" here may be unclear down the line, when other systems may be legacy.

pshiu (approved):
Looks great - thanks so much @MushtaqRossier!

PR #132, opened 2023-11-29
feat: adding `sanctioned` to orders as workflow state

grmartin:
Includes definition, management script and ADR.

This is SONIC-177

### DIFF
--- a/docs/decisions/0007-sdn-sancation-state-on-orders-in-commercetools.rst
+++ b/docs/decisions/0007-sdn-sancation-state-on-orders-in-commercetools.rst
@@ -1,0 +1,41 @@
+7. SDN Sancationed State on Orders in Commercetools for R0.1
+############################################################
+
+Status
+******
+
+**Provisional** *2023-11-29* -- R0.1
+
+Context
+*******
+
+Due to the speed required for R0.1, we needed a way to mark orders as blocked by an SDN match.
+
+This will be implemented by a single `custom State`_ within the Order. This is sometimes referred to as a [Workflow] Transition State.
+
+.. _custom State: https://docs.commercetools.com/api/projects/states
+
+Decision
+********
+
+This has lead to the creation of the following is the CoCo State definition:
+
+.. code-block:: python
+
+    SANCTIONED_ORDER_STATE = StateDraft(
+        key=TwoUKeys.SDN_SANCTIONED_ORDER_STATE,
+        type=StateTypeEnum.ORDER_STATE,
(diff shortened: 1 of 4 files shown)


shafqatfarhan (approved):

docs/decisions/0007-sdn-sancation-state-on-orders-in-commercetools.rst
```diff
@@ -0,0 +1,41 @@
...
+******
+
+**Provisional** *2023-11-29* -- R0.1
+
+Context
+*******
+
+Due to the speed required for R0.1, we needed a way to mark orders as blocked by an SDN match.
+
+This will be implemented by a single `custom State`_ within the Order. This is sometimes referred to as a [Workflow] Transition State.
```
pshiu, line 14: Why not use [PaymentStatus](https://docs.commercetools.com/api/projects/payments#ctp:api:type:PaymentStatus)?

docs/decisions/0007-sdn-sancation-state-on-orders-in-commercetools.rst
```diff
@@ -0,0 +1,41 @@
...
+
+    SANCTIONED_ORDER_STATE = StateDraft(
+        key=TwoUKeys.SDN_SANCTIONED_ORDER_STATE,
+        type=StateTypeEnum.ORDER_STATE,
+        name=ls({'en': 'Sanctioned'}),
+        description=ls({'en': 'This order has been sanctioned for an SDN hit'})
+    )
+
+
+Consequences
```
pshiu, line 33: Do we need to consider confusion between [Order.orderState](https://docs.commercetools.com/api/projects/orders#change-orderstate) and [Order.state](https://docs.commercetools.com/api/projects/orders#transition-state)? How do we intend to use both?

pshiu: For context: Right now, it seems we're proposing the Order.state state machine consist of a single state, and it being the absence of that state that means an order is not sanctioned.

pshiu: Worry: What happens if Support accidentally advances Order.state?

docs/decisions/0007-sdn-sancation-state-on-orders-in-commercetools.rst
pshiu: An Order's workflow state is more searchable in the Merchant Center than a Payment's workflow state.

PR #135, opened 2023-12-13
feat: Redirect to CT based on course availabilty

MushtaqRossier:
**Description**: Redirect course purchasers to the CT checkout page based on CT course availability check

**JIRA**: [SONIC-123](https://2u-internal.atlassian.net/browse/SONIC-123)


commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -28,7 +31,18 @@ def run_filter(self, request):
...
-        if is_redirect_to_commercetools_enabled_for_user(request) and course_run:
+        ct_api_client = CommercetoolsAPIClient()
+        # breakpoint()
+        try:
+            commercetools_available_course =  ct_api_client.get_product_variant_by_course_run(course_run)
+        except HTTPError as exc:
+            logger.exception(
+                '[get_product_variant_by_course_run] Failed to get Commercetools course for course_run: %s', course_run
+            )
+            raise APIException("Error while fetching course variant from Commercetools") from exc
```
shafqatfarhan, line 42: Can we use the legacy system as a fallback option whenever CT SDK throws an error or the product is not found?

```suggestion
        commercetools_available_course = None
        if course_run:
            try:
                commercetools_available_course =  ct_api_client.get_product_variant_by_course_run(course_run)
            except HTTPError as exc:
                logger.exception(
                    '[get_product_variant_by_course_run] Failed to get Commercetools course for course_run: %s', course_run
                )
```

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -218,3 +219,31 @@ def get_orders_for_customer(self, edx_lms_user_id: int, offset=0,
...
+            filter(
+                lambda v: v.sku == cr_id,
+                all_variants
+            )
+        )
+
+        if len(matching_variant_list) < 1:
+            return None
+
+        return matching_variant_list[0]
```
shafqatfarhan, line 249: IMO, we can avoid this further filtering of data to determine whether the product exists or not.

In the case where the product by variant exists, it will return in the `results` object and that should be enough for us to determine the variant's existence

In the case where product by variant does not exist, it will return an empty list `[]`

MushtaqRossier: This was a temporary function to implement in the `GetActiveOrderManagementSystem` pipeline. @grmartin created the file `edx_utils.py` which consists of all the edx helper functions which I will use in the pipeline instead.

pshiu:
@MushtaqRossier: I added a couple of commits, please take a look at their commit messages for more information. With these, my local reports all tests pass. However, you will have to load in [private.in.txt](https://github.com/edx/commerce-coordinator/files/13722999/private.in.txt) to see the pass. I think the remaining work to wrap up the testing errors we were seeing is:
- confirm `make test` with [private.in.txt](https://github.com/edx/commerce-coordinator/files/13722999/private.in.txt) works for you on your local - see comments on top of file for instructions.
- wait for merge of https://github.com/edx/commercetools-python-sdk/pull/1
- commit the results of a `make upgrade`
- in the meantime, bring this branch up-to-date with the base branch (branch 2u/replatforming-lightweight-prototype may be helpful to resolve conflicts)

(fyi: @grmartin)

commerce_coordinator/apps/commercetools/clients.py
MushtaqRossier: The filtering makes it easier to access certain attributes of the product variant object for testing purposes. For example, if not for this filtering, I would not be able to access the sku of the product variant object which is is used in the unit tests.

grmartin: Shafqat, this is also a FT Search... so its using inexact matching (think like soundex) so we need to search the return to ensure an exact match is found... hence the filter.

grmartin (approved):

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -30,7 +34,18 @@ def run_filter(self, request):
...
+        ct_api_client = CommercetoolsAPIClient()
+
+        if course_run:
+            try:
+                commercetools_available_course = ct_api_client.get_product_variant_by_course_run(course_run)
+            except HTTPError as exc:
+                logger.exception(
+                    '[get_product_variant_by_course_run] Failed to get CT course for course_run: %s', course_run
+                )
+                raise APIException("Error while fetching course variant from CT") from exc
```
shafqatfarhan, line 46: This Exception raised is not being caught anywhere in the endpoint from where this pipeline is being initiated.

Ref: https://github.com/edx/commerce-coordinator/blob/main/commerce_coordinator/apps/lms/views.py#L48

shafqatfarhan (approved):

commerce_coordinator/apps/rollout/pipeline.py
MushtaqRossier: Working on this unit test as we speak

grmartin:
Unit Tests have become SONIC-251. Comitting.

PR #155, opened 2024-01-16
feat: Added client that checks for enterprise users

MushtaqRossier:
**Description**: Configured `enterprise_client` that checks whether a user is enterprise or not. Enterprise users will then be redirected to legacy.

**JIRA**: [SONIC-124](https://2u-internal.atlassian.net/browse/SONIC-124)

### DIFF
--- a/commerce_coordinator/apps/enterprise_learner/enterprise_client.py
+++ b/commerce_coordinator/apps/enterprise_learner/enterprise_client.py
@@ -1,0 +1,45 @@
+"""
+API client for calls to Enterprise Service.
+"""
+import logging
+
+from django.conf import settings
+from requests.exceptions import ConnectionError, HTTPError, Timeout  # pylint: disable=redefined-builtin
+
+from commerce_coordinator.apps.core.clients import BaseEdxOAuthClient
+
+logger = logging.getLogger(__name__)
+
+
+class EnterpriseApiClient(BaseEdxOAuthClient):
+    """
+    API client for calls to the Enterprise Service.
+    """
+    enterprise_api_base_url = f"{settings.ENTERPRISE_URL}/enterprise/api/v1"
+    enterprise_customer_info_endpoint = f"{enterprise_api_base_url}/enterprise-learner"
+
+    def check_user_is_enterprise_customer(self, username):
+        """
+        Checks if the user is an enterprise customer.
+        Arguments:
+            username (string): Username of the user
+        Returns:
+            is_enterprise_user (Boolean) : True if the user is an enterprise customer else False
+        """
+        try:
+            response = self.client.get(self.enterprise_customer_info_endpoint,
+                                       timeout=settings.ENTERPRISE_CLIENT_TIMEOUT,
+                                       params={
+                                           'username': username,
+                                       })
+            response.raise_for_status()
+            response_json = response.json()
+        except (ConnectionError, HTTPError, Timeout) as exc:
+            logger.exception(f'An error occurred while requesting enterprise customer data'
+                             f' from  {self.enterprise_customer_info_endpoint}: {exc}')
+            return False
+
+        response_data = response_json.get('count', 0)
+        is_enterprise_user = response_data > 0
+
+        return is_enterprise_user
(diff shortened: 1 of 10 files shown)


commerce_coordinator/settings/base.py
```diff
@@ -90,16 +99,6 @@ def root(*path_fragments):
     'edx_rest_framework_extensions.auth.jwt.middleware.EnsureJWTAuthSettingsMiddleware',
 )
 
-# Cache Configuration
-CACHES = {
```
aht007, line 94: Just to be clear, I removed this configuration for just letting you know about the culprit of the failed test cases. Ideally this configuration shouldn't be picked in the check run that was failing and instead it should use the test file. Also you added [duplicate Cache config](https://github.com/edx/commerce-coordinator/pull/155/files#diff-b2b59c815137417b558b63d9cdafd3965a4b443a0fc911938adcad6541499c7aR65) so you would want to keep just one of them after fixing the configuration file for the check run that was failing.
If we want to use `pymemcache` as a cache backend then we should install it as a dependency too and also make a ticket to configure it for stage and prod

grmartin (approved):

commerce_coordinator/apps/enterprise_learner/enterprise_client.py
```diff
@@ -0,0 +1,45 @@
...
+from commerce_coordinator.apps.core.clients import BaseEdxOAuthClient
+
+logger = logging.getLogger(__name__)
+
+
+class EnterpriseApiClient(BaseEdxOAuthClient):
+    """
+    API client for calls to the Enterprise Service.
+    """
+    enterprise_api_base_url = f"{settings.ENTERPRISE_URL}/enterprise/api/v1"
```
shafqatfarhan, line 18: We can replace `ENTERPRISE_URL` with `LMS_URL_ROOT` since we call the LMS endpoint for enterprise user check. 

If we still intend to use `ENTERPRISE_URL` then we need to define its value in edx-internal for the stage/prod environment.

commerce_coordinator/apps/enterprise_learner/enterprise_client.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Checks if the user is an enterprise customer.
+        Arguments:
+            username (string): Username of the user
+        Returns:
+            is_enterprise_user (Boolean) : True if the user is an enterprise customer else False
+        """
+        try:
+            response = self.client.get(self.enterprise_customer_info_endpoint,
+                                       timeout=settings.ENTERPRISE_CLIENT_TIMEOUT,
```
shafqatfarhan, line 31: Please set this flag's value in edx-internal for stage/prod envs.

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -47,7 +48,7 @@ def run_filter(self, request):
 
         if is_redirect_to_commercetools_enabled_for_user(request) and commercetools_available_course:
             active_order_management_system = COMMERCETOOLS_FRONTEND
-        elif sku:
+        elif sku and is_user_enterprise_learner(request):
             active_order_management_system = FRONTEND_APP_PAYMENT_CHECKOUT
```
shafqatfarhan, line 52: We want to disallow enterprise learners to purchase via CT. The current implementation logic is doing otherwise. It is not checking for enterprise learners in the condition on which we are setting CT as the active order management system.

The logic should be like:

```  
if is_redirect_to_commercetools_enabled_for_user(request) and commercetools_available_course and not is_user_enterprise_learner(request):
    active_order_management_system = COMMERCETOOLS_FRONTEND
elif sku:
    active_order_management_system = FRONTEND_APP_PAYMENT_CHECKOUT
```

shafqatfarhan (commented):
@grmartin @MushtaqRossier @pshiu  I was testing this feature on staging and found out it is not working as intended on staging. Turns out this is merged in the `main` branch and we need to merge these changes in `2u/sandbox-lightweight-prototype` branch which is deployed on staging env.
 
Also, I looked into the implementation and found a few improvements. Please have a look.

commerce_coordinator/apps/rollout/pipeline.py
MushtaqRossier: @shafqatfarhan I can add these changes as part of [SONIC-251](https://2u-internal.atlassian.net/browse/SONIC-251)

commerce_coordinator/apps/enterprise_learner/enterprise_client.py
MushtaqRossier: Just merged the flag's value to edx-internal

PR #159, opened 2024-01-24
feat: Modified CT order history implementation

JadeyOlivier:
Description
Verified the current implementation to retrieve the customers order history from Commercetools is functioning correctly and have modified the implementation to only fetch completed orders.

JIRA
https://2u-internal.atlassian.net/browse/SONIC-128

### DIFF
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -170,7 +170,8 @@ class CommercetoolsAPIClient:
             return results.results[0]
 
     def get_orders(self, customer: CTCustomer, offset=0,
-                   limit=ORDER_HISTORY_PER_SYSTEM_REQ_LIMIT) -> PaginatedResult[CTOrder]:
+                   limit=ORDER_HISTORY_PER_SYSTEM_REQ_LIMIT,
+                   order_state="Complete") -> PaginatedResult[CTOrder]:
 
         """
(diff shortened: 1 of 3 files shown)


grmartin (approved):
Fix the null check but otherwise you have my approval :)

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -170,7 +170,8 @@ def get_customer_by_lms_user_id(self, lms_user_id: int) -> Optional[CTCustomer]:
             return results.results[0]
 
     def get_orders(self, customer: CTCustomer, offset=0,
-                   limit=ORDER_HISTORY_PER_SYSTEM_REQ_LIMIT) -> PaginatedResult[CTOrder]:
+                   limit=ORDER_HISTORY_PER_SYSTEM_REQ_LIMIT,
+                   where="orderState=\"Complete\"") -> PaginatedResult[CTOrder]:
```
aht007, line 174: If we are making the order state a param then I would suggest we instead just pass `order_state` param with default value of `Complete` and format it inside the `get_orders` function. This would make calling this utility easier on the part of user and we can do the remaining formatting inside the utility.

aht007 (changes requested):
Left a comment, Also the coverage is failing for which you would need to ensure the code is covered in tests

aht007 (approved):
LGTM

PR #163, opened 2024-01-31
feat: Modified CT order history implementation (#159)

pshiu:
Replays fd5609a on main.


grmartin (approved):
This looks like Jade's work ive already approve go for it

PR #165, opened 2024-02-06
fix: Ecomm starts pagination like a book not a database...

grmartin:
The old 1 vs 0 issue.

SONIC-126

### DIFF
--- a/commerce_coordinator/apps/ecommerce/pipeline.py
+++ b/commerce_coordinator/apps/ecommerce/pipeline.py
@@ -21,7 +21,9 @@ class GetEcommerceOrders(PipelineStep):
         """
 
         ecommerce_api_client = EcommerceAPIClient()
-        ecommerce_response = ecommerce_api_client.get_orders(params)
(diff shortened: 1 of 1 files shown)


shafqatfarhan (approved):

commerce_coordinator/apps/ecommerce/pipeline.py
```diff
@@ -21,7 +21,9 @@ def run_filter(self, params, order_data):  # pylint: disable=arguments-differ
         """
 
         ecommerce_api_client = EcommerceAPIClient()
-        ecommerce_response = ecommerce_api_client.get_orders(params)
+        new_params = params.copy()
+        new_params['page'] = params['page'] + 1
```
grmartin, line 25: We currently ignore the input page number and set it to 0, thus for the moment... (while Pagination is disabled) we have to adjust Ecomm's version of this value. We may have to reverse and do so in the CT pipeline while implementing pagination or something else.

PR #171, opened 2024-02-13
SONIC-279: Receipt URLs and Generation

grmartin:


grmartin:
@MushtaqRossier Can we meet Wednesday to go over this test, i cant find what i broke to break this.

commerce_coordinator/apps/ecommerce/constarnts.py
```diff
@@ -0,0 +1,2 @@
+""" Ecommerce Application Consonants """
```
shafqatfarhan, line 1: typo in the file name: `constants.py`

```suggestion
""" Ecommerce Application Constants """
```

commerce_coordinator/apps/ecommerce/pipeline.py
```diff
@@ -2,9 +2,26 @@
...
+from commerce_coordinator.apps.ecommerce.constarnts import ECOMMERCE_ORDER_MANAGEMENT_SYSTEM
+
+
+class GetLegacyEcommerceReceiptRedirectUrl(PipelineStep):
+    """ Returns a redirect URL for an Ecommerce order if this is the active system. """
+
+    def run_filter(self, params, active_order_management_system, order_number):
+        # pylint: disable=arguments-differ, unused-argument
+        if not active_order_management_system == ECOMMERCE_ORDER_MANAGEMENT_SYSTEM and \
+                active_order_management_system is not None:
```
shafqatfarhan, line 19:
```suggestion
        if active_order_management_system != ECOMMERCE_ORDER_MANAGEMENT_SYSTEM:
```

commerce_coordinator/settings/base.py
```diff
@@ -345,7 +345,7 @@ def root(*path_fragments):
 FULFILLMENT_TIMEOUT = 7
 
 # API URLs
-ECOMMERCE_URL = 'https://ecommerce_url/'
+ECOMMERCE_URL = 'http://localhost:18130/'
```
shafqatfarhan, line 348: To match it with the value set in edx-internal, the URL base should not have a trailing slash.

```suggestion
ECOMMERCE_URL = 'http://localhost:18130'
```

commerce_coordinator/settings/base.py
```diff
@@ -477,7 +486,11 @@ def root(*path_fragments):
     'scopes': 'some_scope'
 }
 
-# Setting to keep using deperecated pytz with Django>4
+
+# Will be suffixed with order numbers
+ECOMMERCE_RECEIPT_URL_BASE = f'{ECOMMERCE_URL}checkout/receipt/?order_number='
```
shafqatfarhan, line 491:
```suggestion
ECOMMERCE_RECEIPT_URL_BASE = f'{ECOMMERCE_URL}/checkout/receipt/?order_number='
```

commerce_coordinator/apps/rollout/utils.py
```diff
@@ -0,0 +1,17 @@
+""" Rollout Utility Functions"""
+
+import re
+
+
+def is_legacy_order(order_number: str) -> bool:
+    """ Determine if an order is a Legacy Ecommerce order """
+    if not order_number:
+        return False
```
shafqatfarhan, line 9: IMO, this check should be added to the `is_legacy_order`'s caller code block.

commerce_coordinator/apps/rollout/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    return re.search("EDX-[0-9]{1,6}", order_number) is not None
+
+
+def is_uuid(value: str) -> bool:
+    """ Determine if an value is a UUID """
+    if not value:
+        return False
+    return re.search(r"\b[A-Fa-f0-9]{8}(?:-[A-Fa-f0-9]{4}){3}-[A-Fa-f0-9]{12}\b", value) is not None
```
shafqatfarhan, line 17: We can check for a valid UUID by initializing its object:

```suggestion
    from uuid import UUID
    is_valid = True
    try:
        uuid_obj = UUID(uuid_to_test)
    except ValueError:
        is_valid=False
    return is_valid
```

commerce_coordinator/apps/rollout/pipeline.py
```diff
@@ -51,11 +59,42 @@ def run_filter(self, request):
...
+
+    def run_filter(self, order_number, **kwargs):  # pylint: disable=arguments-differ
+        """ Using an Order Numer/ID to determine the active order management system """
+
+        active_order_management_system = None
+
+        if is_uuid(order_number):
+            active_order_management_system = COMMERCETOOLS_ORDER_MANAGEMENT_SYSTEM
+        elif is_legacy_order(order_number):
+            active_order_management_system = ECOMMERCE_ORDER_MANAGEMENT_SYSTEM
```
shafqatfarhan, line 83: Can we only check for `is_legacy_order()` and fallback to CT order always? The reason is that we know our legacy order number starts with `EDX-...`. Also, in the future, we plan to move away from CT Order's UUID and will be using human-readable order_number.

commerce_coordinator/apps/frontend_app_ecommerce/views.py
```diff
@@ -25,6 +30,41 @@ def date_conv(dt: Union[datetime, str]) -> datetime:
...
+            "order_number": request.query_params.get('order_number', None),
+        }
+
+        # deny global queries
+        if not request.user.username:  # pragma: no cover
+            # According to the Django checks, this isn't possible with our current user model.
+            # Leaving in incase that changes.
+            raise PermissionDenied(detail="Could not detect username.")
+        if not request.user.lms_user_id:  # pragma: no cover
+            raise PermissionDenied(detail="Could not detect LMS user id.")
```
shafqatfarhan, line 55: IMO, we don't need these checks here. Our JWT authentication ensures to creation of a new user in the service and mapp its LMS ID and username.

commerce_coordinator/apps/frontend_app_ecommerce/urls.py
```diff
@@ -4,10 +4,11 @@
 
 from django.urls import include, path
 
-from commerce_coordinator.apps.frontend_app_ecommerce.views import UserOrdersView
+from commerce_coordinator.apps.frontend_app_ecommerce.views import RedirectReceiptView, UserOrdersView
 
 app_name = 'frontend_app_ecommerce'
 urlpatterns = [
     path('order_history/', UserOrdersView.as_view(), name='order_history'),
+    path('order_receipt', RedirectReceiptView.as_view(), name='order_receipt'),
```
shafqatfarhan, line 12:
```suggestion
    path('order_receipt/', RedirectReceiptView.as_view(), name='order_receipt'),
```

commerce_coordinator/apps/frontend_app_ecommerce/views.py
```diff
@@ -25,6 +30,41 @@ def date_conv(dt: Union[datetime, str]) -> datetime:
...
+    throttle_classes = [UserRateThrottle]
+
+    def get(self, request):
+        """Get the order history for the authenticated user."""
+
+        # build parameters
+        params = {
+            'username': request.user.username,
+            "edx_lms_user_id": request.user.lms_user_id,
+            "order_number": request.query_params.get('order_number', None),
```
shafqatfarhan, line 46: We should check here if the order_number is None and return a 404 response. This will eventually not execute pipeline filters and would save us some processing logic.

commerce_coordinator/apps/frontend_app_ecommerce/filters.py
```diff
@@ -4,6 +4,29 @@
...
+class OrderReceiptRedirectionUrlRequested(OpenEdxPublicFilter):
+    """
+    Filter to gather order receipt redirection url from the defined PipelineStep(s)
+    """
+
+    # See pipeline step configuration OPEN_EDX_FILTERS_CONFIG dict in `settings/base.py`
+    filter_type = "org.edx.coordinator.frontend_app_ecommerce.order.receipt_url.requested.v1"
+
+    @classmethod
+    def run_filter(cls, params, order_number=None):
```
shafqatfarhan, line 16:
```suggestion
    If we ensure that order_number is always present before executing this pipeline, we do not need to provide a default value here.
    
    def run_filter(cls, params, order_number):
```

shafqatfarhan (commented):
Left some comments. Please have a look, thanks.

commerce_coordinator/apps/frontend_app_ecommerce/urls.py
grmartin: Why? It works fine without it? and the other project is coded to not have it?

commerce_coordinator/apps/rollout/pipeline.py
grmartin: I was thinking we may have other systems in the future... so i was being thorough... but sure.

commerce_coordinator/apps/rollout/utils.py
grmartin: This seemed like more work... especially since i had the regex handy...

commerce_coordinator/apps/frontend_app_ecommerce/views.py
grmartin: We were doing this in other places... So i figured following the previous pattern was more correct.

commerce_coordinator/apps/rollout/utils.py
grmartin: Utility functions usually assume a caller can be fallible.

grmartin:
Related: 
https://github.com/openedx/frontend-app-ecommerce/pull/371
https://github.com/edx/edx-internal/pull/10177/files

commerce_coordinator/apps/frontend_app_ecommerce/urls.py
grmartin: Answer: trailing slash is what nginx appends to URLs during routing to the appropriate service. If it is not present, no endpoint would match

shafqatfarhan (approved):

PR #172, opened 2024-02-16
Add: Update ReturnPaymentState on Order

JadeyOlivier:
### Description: 
Added Commercetools Client method to update the return payment state for an order to be called after a successful stripe refund. Started to configure pipeline for Returns on Commercetools orders.

### JIRA: 
[SONIC-92](https://2u-internal.atlassian.net/browse/SONIC-92)


commerce_coordinator/apps/commercetools/tests/_test_cases.py
pshiu, line 1: I'm not quite sure what is happening here. Is it possible to make this more concise, à la:

https://github.com/edx/commerce-coordinator/blob/a932b18f5ef250807c06663f5da9f8673678005f/commerce_coordinator/apps/titan/tests/test_pipeline.py#L512-L521

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -269,3 +272,39 @@ def get_product_variant_by_course_run(self, cr_id: str) -> Optional[CTProductVar
...
+            )
+
+            updated_order = self.base_client.orders.update_by_id(
+                id=order_id,
+                version=order_version,
+                actions=[return_payment_state_action]
+            )
+            return updated_order
+        except CommercetoolsError as err:
+            error_info = {
```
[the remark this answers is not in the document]
JadeyOlivier: I suppose we could - had split it out initially for easier readability, but if you feel we would get the same value from the `err.errors` directly, we can change this. Based on your previous comment, would we then print out the `err.errors` and the `err.correlation_id` in the output string?

commerce_coordinator/apps/commercetools/tests/test_clients.py
```diff
@@ -308,6 +314,77 @@ def test_order_history_with_limits(self):
...
+                    )
+
+                exception = cm.exception
+
+                expected_message = (
+                    f"[CommercetoolsError] Unable to update ReturnPaymentState of order "
+                    f"mock_order_id with error correlation id {exception.correlation_id} "
+                    f"and error/s: {exception.errors}"
+                )
+                pprint_mock.assert_called_once_with(expected_message)
```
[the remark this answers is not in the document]
pshiu: Yeah, sorry for misleading you here on needing `pprint`, @JadeyOlivier - I agree with Glenn, think we can just use the logger.

commerce_coordinator/apps/commercetools/pipeline.py
```diff
@@ -38,3 +39,30 @@ def run_filter(self, params, order_data):  # pylint: disable=arguments-differ
         return {
             "order_data": order_data
         }
+
+
+class UpdateCommercetoolsOrderReturnPaymentStatus(PipelineStep):
```
aht007, line 44: Is there a separate ticket for utilizing this pipeline step?

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -269,3 +271,36 @@ def get_product_variant_by_course_run(self, cr_id: str) -> Optional[CTProductVar
...
+        Returns Exception: Error if update was unsuccesful.
+        """
+        try:
+
+            return_payment_state_action = OrderSetReturnPaymentStateAction(
+                return_item_id=return_item_id,
+                payment_state=ReturnPaymentState.REFUNDED
+            )
+
+            updated_order = self.base_client.orders.update_by_id(
```
aht007, line 296: Have you verified what happens if send this update call again for an order which is already refunded?

commerce_coordinator/apps/commercetools/pipeline.py
JadeyOlivier: Yes, that will be done as part of a separate ticket to integrate all of the automatic refunds work together.

commerce_coordinator/apps/commercetools/clients.py
JadeyOlivier: No I have not been able to as this is the last step in the automatic refunds process and as such made testing this tricky. 
I could map through the order's return item array to find the return item we wish to update and do a check for it's current value to prevent us from updating if that value is already set to refunded?

JadeyOlivier: I agree. The CommercetoolsError exception should catch this case

shafqatfarhan (approved):
👍

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -269,3 +271,36 @@ def get_product_variant_by_course_run(self, cr_id: str) -> Optional[CTProductVar
...
+                id=order_id,
+                version=order_version,
+                actions=[return_payment_state_action]
+            )
+            return updated_order
+        except CommercetoolsError as err:
+            logger.error(f"[CommercetoolsError] Unable to update ReturnPaymentState "
+                         f"of order {order_id} with error correlation id {err.correlation_id} "
+                         f"and error/s: {err.errors}")
+            raise
```
shafqatfarhan, line 306:
```suggestion
            raise err
```

grmartin (approved):

PR #182, opened 2024-03-08
feat: Transition LineItemState on fulfillment

JadeyOlivier:
### Description
Created new LineItemStates in Commercetools with accompanying transitions to track fulfillment process. Coordinator triggers fulfillment on “Fulfillment Pending” events when a Commercetools message is received from the AWS Eventbridge. Based on fulfillment result, the LineItemState is also transitioned to a success or failure state. Initially transitioning the state to 'Fulfillment Pending' is handle in this [PR](https://github.com/frontastic-developers/customer-twou/pull/98/) in the Commercetools twou repo

### JIRA
[SONIC-270](https://2u-internal.atlassian.net/browse/SONIC-270)

### DIFF
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -12,9 +12,11 @@ from commercetools.platform.models import Customer as CTCustomer
 from commercetools.platform.models import CustomerSetCustomTypeAction as CTCustomerSetCustomTypeAction
 from commercetools.platform.models import FieldContainer as CTFieldContainer
 from commercetools.platform.models import Order as CTOrder
-from commercetools.platform.models import OrderAddReturnInfoAction, OrderSetReturnPaymentStateAction
-from commercetools.platform.models import OrderTransitionLineItemStateAction
-from commercetools.platform.models import LineItem as CTLineItem
+from commercetools.platform.models import (
+  OrderAddReturnInfoAction,
+  OrderSetReturnPaymentStateAction,
+  OrderTransitionLineItemStateAction
+)
 from commercetools.platform.models import ProductVariant as CTProductVariant
 from commercetools.platform.models import ReturnItemDraft, ReturnPaymentState, ReturnShipmentState
 from commercetools.platform.models import Type as CTType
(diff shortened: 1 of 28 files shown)


commerce_coordinator/apps/commercetools/views.py
```diff
@@ -26,12 +28,13 @@
...
 )
 from .clients import CommercetoolsAPIClient
-from .serializers import OrderMessageInputSerializer
+from .serializers import OrderMessageInputSerializer, OrderLineItemMessageInputSerializer
 from .utils import (
     extract_ct_order_information_for_braze_canvas,
     extract_ct_product_information_for_braze_canvas,
     send_order_confirmation_email
 )
+from commerce_coordinator.apps.core.signal_helpers import format_signal_results
```
JadeyOlivier, line 37: This unused import will be removed in an upcoming commit

commerce_coordinator/apps/commercetools/views.py
```diff
@@ -93,21 +98,47 @@ def post(self, request):
         canvas_entry_properties.update(extract_ct_order_information_for_braze_canvas(customer, order))
 
         for item in get_edx_items(order):
-            logger.debug('[CT-OrderFulfillView] processing edX order %s, line item %s', order_id, item.variant.sku)
+            logger.debug('[CT-OrderFulfillView] processing edX order %s, line item %s', order.id, item.variant.sku)
+
+            state_ids = []
+            for item_state in item.state:
```
JadeyOlivier, line 104: Based on the outcome of [this discussion](https://github.com/frontastic-developers/customer-twou/pull/98/files#r1539093936), we may not need to iterate through all the line item states and instead would want to only update the first state in the array. This may change in an upcoming commit.

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -240,6 +258,12 @@ def get_orders_for_customer(self, edx_lms_user_id: int, offset=0,
     def get_customer_by_id(self, customer_id: str) -> CTCustomer:
         return self.base_client.customers.get_by_id(customer_id)  # pragma no cover
 
+    def get_state_by_id(self, state_id: str) -> CTLineItemState:
+        return self.base_client.states.get_by_id(state_id)  # pragma no cover
```
grmartin, line 262: We should test these in the future.

grmartin (approved):

PR #185, opened 2024-03-15
fix: matching f-a-ecomms returns exactly

grmartin:
- Added settings for future formatting
- Updating local settings to use the provisioned key name, even though it fails the standard format by having an underscore
- Updating devstack to not use sqlite

### DIFF
--- a/commerce_coordinator/apps/commercetools/catalog_info/utils.py
+++ b/commerce_coordinator/apps/commercetools/catalog_info/utils.py
@@ -12,7 +12,7 @@ from currencies import Currency
 from commerce_coordinator.apps.commercetools.catalog_info.constants import (
     HIDE_CODE_FOR_CURRENCIES,
     LS_OUT_PREFERENCES,
-    SEND_MONEY_AS_DECIMAL,
+    SEND_MONEY_AS_DECIMAL_STRING,
     Languages
 )
 
@@ -74,7 +74,7 @@ def un_ls(string_dict: LSLike, preferred_lang: Optional[str] = None):
         return string_dict[langs_available[0]]
 
 
-def price_to_string(price: CTPrice, money_as_decimal=False) -> str:
+def price_to_string(price: CTPrice, money_as_decimal_string=False) -> str:
     """
     Convert Commercetools price to a string.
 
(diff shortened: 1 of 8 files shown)


commerce_coordinator/settings/local.py
```diff
@@ -78,11 +78,11 @@
 )
 BACKEND_SERVICE_EDX_OAUTH2_KEY = os.environ.get(
     'BACKEND_SERVICE_EDX_OAUTH2_KEY',
-    'commerce-coordinator-backend-service-key'
+    'commerce_coordinator-backend-service-key'
 )
 BACKEND_SERVICE_EDX_OAUTH2_SECRET = os.environ.get(
     'BACKEND_SERVICE_EDX_OAUTH2_SECRET',
-    'commerce-coordinator-backend-service-secret'
+    'commerce_coordinator-backend-service-secret'
```
shafqatfarhan, line 85: We run the commerce coordinator's service locally using the `local.py` settings file as mentioned in the [REAMDE](https://github.com/edx/commerce-coordinator?tab=readme-ov-file#local). Would this change affect all the existing developers who have already run the provisioning script that creates the service user locally using this OAUTH key and secret? (Ref: https://github.com/edx/commerce-coordinator/blob/f0570c3a6d4b0fe1671c207872e2096d44f121a3/provision-commerce-coordinator.sh#L54)

commerce_coordinator/apps/commercetools/tests/test_data.py
```diff
@@ -127,8 +127,8 @@ def test_convert_line_item(self):
                 course_organization="",
                 description=un_ls(li.name),
                 status="PAID",
-                line_price_excl_tax=price_to_string(li.price),
-                unit_price_excl_tax=price_to_string(li.price)
+                line_price_excl_tax=price_to_string(li.price, SEND_MONEY_AS_DECIMAL),
```
shafqatfarhan, line 130: nit: named argument
```suggestion
                line_price_excl_tax=price_to_string(li.price, money_as_decimal=SEND_MONEY_AS_DECIMAL),
```

commerce_coordinator/apps/commercetools/data.py
```diff
@@ -131,7 +131,8 @@ def order_from_commercetools(order: CTOrder, customer: CTCustomer) -> LegacyOrde
                     order.discount_on_total_price.discounted_amount if order.discount_on_total_price else None
                 ),
                 order.taxed_price.total_tax
-            )
+            ),
+            SEND_MONEY_AS_DECIMAL
```
shafqatfarhan, line 135: nit: named argument
```suggestion
            money_as_decimal=SEND_MONEY_AS_DECIMAL
```

commerce_coordinator/apps/commercetools/catalog_info/utils.py
```diff
@@ -103,6 +106,10 @@ def typed_money_to_string(money: CTTypedMoney) -> str:
     cur = Currency(money.currency_code)
 
     def _format(cost):
+        if money_as_decimal:  # pragma no cover
+            # This is used to support legacy ecomm mfe
+            return str(cost)
```
shafqatfarhan, line 111: nit: The variable name reads as money in decimal type. But we are type-casting it to a string here.

shafqatfarhan (approved):
Left a couple of comments. Otherwise, looks good.

PR #166, opened 2024-02-08
fix: restrict commercetools /fulfill view to staff users

pshiu:
## Description

Add tests and permission flags for /commercetools/fulfill view to be restricted only to staff users.

## Additional Information

* Jira: [SONIC-268](https://2u-internal.atlassian.net/browse/SONIC-268)

PR #188, opened 2024-03-18
feat: add authentication for commercetools views

pshiu:
## Description

The views for the Commercetools app are implemented to allow an external service configured with a Commercetools Subscription, like AWS EventBridge, to send Commercetools Messages to Coordinator.

AWS EventBridge supports retrieving a token from an OAuth2 service. In this case, we have instructed it to get a token from LMS. However, AWS EventBridge will expect that the consuming API Destination perform authorization by expecting a header like:

    Authorization: Bearer <token>

Coordinator is configured to expect headers like:

    Authorization: JWT <token>

Because this is what other MFEs use.

This commit adds class `JwtBearerAuthentication` to allow Coordinator to expect the new header format, as well as enforce that authentication method on existing views in Coordinator's Commercetools app.

## Additional Information

* Jira: [SONIC-268](https://2u-internal.atlassian.net/browse/SONIC-268) Set up authentication between EventBridge & Coordinator
* Reapplies / un-reverts: https://github.com/edx/commerce-coordinator/pull/174
* Follow on to: https://github.com/edx/commerce-coordinator/pull/166

### DIFF
--- a/commerce_coordinator/apps/commercetools/views.py
+++ b/commerce_coordinator/apps/commercetools/views.py
@@ -5,6 +5,7 @@ import logging
 
 from commercetools import CommercetoolsError
 from rest_framework import status
+from rest_framework.authentication import SessionAuthentication
 from rest_framework.permissions import IsAdminUser
 from rest_framework.response import Response
 from rest_framework.views import APIView
@@ -13,6 +14,7 @@ from commerce_coordinator.apps.commercetools.catalog_info.constants import TwoUK
 from commerce_coordinator.apps.commercetools.serializers import OrderFulfillViewInputSerializer
 from commerce_coordinator.apps.commercetools.signals import fulfill_order_placed_signal
 
+from .authentication import JwtBearerAuthentication
 from .catalog_info.edx_utils import (
     get_edx_is_sanctioned,
     get_edx_items,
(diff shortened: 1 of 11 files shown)


shafqatfarhan (approved):
Looks good! We intend to have an endpoint for Order refunds as well which will also require us to set the same `authentication_classes` list. Not sure if it is WIP or in the backlog.

commerce_coordinator/apps/commercetools/views.py
```diff
@@ -39,6 +41,9 @@
 class OrderFulfillView(APIView):
     """Order Fulfillment View"""
 
+    authentication_classes = [JwtBearerAuthentication, SessionAuthentication]
```
shafqatfarhan, line 44: nit: Just setting the `JwtBearerAuthentication` class is required here.
```suggestion
    authentication_classes = (JwtBearerAuthentication,)
```

commerce_coordinator/apps/commercetools/views.py
```diff
@@ -113,6 +118,8 @@ def post(self, request):
 # noinspection DuplicatedCode
 class OrderSanctionedView(APIView):
     """View to sanction an order and deactivate the lms user"""
+
+    authentication_classes = [JwtBearerAuthentication, SessionAuthentication]
```
shafqatfarhan, line 122: nit: Just setting the `JwtBearerAuthentication` class is required here.
```suggestion
    authentication_classes = (JwtBearerAuthentication,)
```

grmartin (approved):

commerce_coordinator/apps/commercetools/views.py
pshiu: I tried using only `JwtBearerAuthentication`, but `SessionAuthentication` is required for tests to pass.

commerce_coordinator/apps/commercetools/views.py
pshiu: I tried using only `JwtBearerAuthentication`, but `SessionAuthentication` is required for tests to pass.

PR #196, opened 2024-04-03
feat: add lms.views.RefundView

pshiu:
**Merge checklist:**
Check off if complete *or* not applicable:
- [ ] Documentation updated (not only docstrings)
- [ ] Fixup commits are squashed away
- [ ] Unit tests added/updated
- [ ] Manual testing instructions provided
- [ ] Noted any: Concerns, dependencies, migration issues, deadlines, tickets

**Post-merge:**
- [ ] [Backported](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2065367719/Making+a+pull+request+for+a+named+release) to latest and next named releases

PR #204, opened 2024-04-18
fix: Rework of Coordinator Signals and Provisioning and standardized Dates

grmartin:
This is based on Phil's PR: https://github.com/edx/commerce-coordinator/pull/196


commerce_coordinator/settings/test.py
```diff
@@ -40,6 +40,145 @@
     }
 }
 
+CC_SIGNALS = {
+    # DEMO: This configuration is just for proof-of-concept and can
+    # be removed once we have real signals
+    'commerce_coordinator.apps.core.signals.test_signal': [
```
JadeyOlivier, line 46: If we are removing these configurations from base, why are we adding them here? Are we simply ensuring references to these signals in the rest of the codebase have somewhere to point to to reduce errors, until we've cleaned up these signals in their respective apps? Or was this missed?

Is there also a ticket for future work to refactor the codebase to remove these unnecessary filters and signal functions?

JadeyOlivier (approved):
LGTM overall 🔥 
Let a small comment

commerce_coordinator/settings/test.py
grmartin: Code coverage drops if the code is there and not tested... We would go from like 99% to like 40% coverage... we can guy them after release when we remove all the titan work.

grmartin: There isnt a ticket, ill make one.

PR #207, opened 2024-04-19
feat: modify order history to use human readable number

nyujacky:


nyujacky:
Not sure if the call to [get order](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/commercetools/pipeline.py#L85) should be changed as well, to` get_order_by_number(order_number=order_number)` since it currently uses `get_order_by_id(order_id=order_number)` where order number is cart-id-### (old order number)

commerce_coordinator/apps/commercetools/data.py
```diff
@@ -111,7 +111,7 @@ def order_from_commercetools(order: CTOrder, customer: CTCustomer) -> LegacyOrde
         date_placed=order.last_modified_at,
         total_excl_tax=typed_money_to_string(order.total_price, money_as_decimal_string=SEND_MONEY_AS_DECIMAL_STRING),
         # in dev systems, this isn't set... so let's use UUID, otherwise, let's rely on order number
-        number=order.id,  # Long-term: this_or(order.order_number, order.id),
+        number=order.order_number or order.id,
```
shafqatfarhan, line 114: Please remove the comment above.
```suggestion
        number=order.order_number,
```

shafqatfarhan, line 114: Since we will be passing and displaying order_number for the order history page, we need to make a change in CC to rely on the order number for the Order receipt request.

Need to change the method to [get_order_by_number](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/commercetools/clients.py#L195)(): https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/commercetools/pipeline.py#L85

shafqatfarhan (commented):
Left a few comments here and over [Slack](https://twou.slack.com/archives/C065NLHU1NF/p1713893649663619?thread_ts=1713824375.152509&cid=C065NLHU1NF) as well. Please have a look. Thanks.

commerce_coordinator/apps/commercetools/pipeline.py
```diff
@@ -111,15 +110,14 @@ class CreateReturnForCommercetoolsOrder(PipelineStep):
...
             active_order_management_system,
-            order_number,
+            order_id,
             order_line_id
     ):  # pylint: disable=arguments-differ
         """
         Execute a filter with the signature specified.
         Args:
             active_order_management_system: The Active Order System
             order_number: Order number (for now this is an order.id, but this should change in the future)
```
shafqatfarhan, line 120: Need to update this argument in the docstring.

```suggestion
            order_id: Order ID
```

commerce_coordinator/apps/lms/filters.py
```diff
@@ -33,17 +33,15 @@ class OrderRefundRequested(OpenEdxPublicFilter):
...
-            order_number: Order number (for now this is an order.id, but this should change in the future)
-            TODO: SONIC-277 (in-progress)
+            order_id: Order Id
             order_line_id: ID of order line item
         Returns:
             order_refund: Updated order with return item attached
         """
 
-        order_refund = super().run_pipeline(order_number=order_number, order_line_id=order_line_id)
+        order_refund = super().run_pipeline(order_id=order_id, order_line_id=order_line_id)
```
shafqatfarhan, line 46:
```suggestion
        order_refund = super().run_pipeline(order_number=order_id, order_line_id=order_line_id)
```
Argument naming inconsistency between pipeline steps:

If we look at the pipeline steps [here](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/settings/base.py#L377-L385), we are calling 2 pipelines and first step's argument name is [order_number](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/rollout/pipeline.py#L76).

And, in our second pipeline step, we have renamed the argument to `order_id` in this PR. I am not sure if this pipeline filter can pass the argument correctly and work properly. You might need to test this.

commerce_coordinator/apps/commercetools/pipeline.py
nyujacky: I think I misunderstood. The intention was to eventually replace the order.id with an actual order number, not to replace the order number argument with order id.

commerce_coordinator/apps/lms/filters.py
nyujacky: I've renamed the second pipeline step back to order_number, for consistency and because I misinterpreted the pipeline argument code comment. It is now back to order_number

shafqatfarhan: It seems like you may not have a clear understanding of the code flow. Let me explain how this pipeline works:

- [OrderRefundRequested](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/lms/views.py#L225) gets called in our `RefundView` endpoint. It takes two arguments, `order_id` and `order_line_id`. These are arguments that were originally passed from LMS's enrollment attributes table. `order_id` argument has a value of CT's order's ID (UUID), not the human-readable order number

- If we look at the pipeline steps [here](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/settings/base.py#L377-L385), we are calling 2 pipelines

- first pipeline [step](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/rollout/pipeline.py#L76) has its first argument named as `order_number` (I am not sure the reason, but it is used in multiple pipelines) but it stores the CT Order's ID (UUID) as a value. This is where all the confusion starts from. We should either change its argument name to `order_id` and test all the filters where this pipeline is used to ensure that it is working fine or create a new pipeline step for this view for the same purpose and pass `order_id` as the argument name

- For the above reason, I suggested to keep the [run_pipline()](https://github.com/edx/commerce-coordinator/pull/207/files?diff=unified&w=1#r1580658992) function's argument name consistent `order_number`

- second pipeline [step](https://github.com/edx/commerce-coordinator/blob/ca054ba93b3d612ea740b5b4868ae96adafbc9c7/commerce_coordinator/apps/commercetools/pipeline.py#L114) has arguments passed from previous (first pipeline) and the argument `order_number` it has stores the CT Order's ID (UUID) as a value. This is the value passed to it from the view from where this pipeline filter is called from

I hope you have got the code flow understanding now. LMK, if you need to sync on this.

commerce_coordinator/apps/commercetools/pipeline.py
```diff
@@ -131,7 +129,7 @@ def run_filter(
 
         try:
             ct_api_client = CommercetoolsAPIClient()
-            order = ct_api_client.get_order_by_id(order_id=order_number)
+            order = ct_api_client.get_order_by_number(order_number=order_number)
```
shafqatfarhan, line 132: Please see #5 on this [comment](https://github.com/edx/commerce-coordinator/pull/207#discussion_r1582752172)

commerce_coordinator/apps/commercetools/pipeline.py
```diff
@@ -70,25 +69,82 @@ def run_filter(self, params, order_data):  # pylint: disable=arguments-differ
...
+            return ret_val
+        except CommercetoolsError as err:  # pragma no cover
+            log.exception(f"[{type(self).__name__}] Commercetools Error: {err}, {err.errors}")
+            return PipelineCommand.CONTINUE.value
+        except HTTPError as err:
+            log.exception(f"[{type(self).__name__}] HTTP Error: {err}")
+            return PipelineCommand.CONTINUE.value
+
+
+class FetchOrderDetailsByOrderID(PipelineStep):
```
grmartin, line 123: Wasnt the idea to use order number everywhere thus gutting all use of order ID as input? We may still need it for lower logic but it can be provided by fetching the order by number and adding it to the pipe as a return.

grmartin (changes requested):

grmartin (approved):

PR #209, opened 2024-04-24
feat: Adding Stripe Refund.

grmartin:


commerce_coordinator/apps/stripe/pipeline.py
```diff
@@ -231,3 +234,44 @@ def run_filter(self, payment_intent_id=None, **params):
...
+
+        stripe_api_client = StripeAPIClient()
+
+        try:
+            ret_val = stripe_api_client.refund_payment_intent(
+                payment_intent_id=payment_intent_id,
+                amount=amount_in_cents,
+            )
+            return {
+                'refund_response': ret_val
```
JadeyOlivier, line 274: According to the [client's](https://github.com/edx/commerce-coordinator/blob/main/commerce_coordinator/apps/stripe/clients.py#L307) return we will only get the id of the refund object back. Could we modify the client return to send the entire refund object back so the next step in the pipeline (which would be creating the CT transaction based on the stripe refund) has access to all the refunds attributes?

grmartin: I dont see why not... ill do that in my work in a moment... since i havent built the stripe tests.

commerce_coordinator/apps/commercetools/tests/test_pipeline.py
```diff
@@ -64,7 +64,8 @@ def test_legacy_ecommerce_refund(self):
         ret = refund_pipe.run_filter(
             active_order_management_system="Legacy",
             order_number="mock_id",
-            order_line_id="mock_line_id"
+            order_line_id="mock_line_id",
+                order_data=self.mock_response_order
```
JadeyOlivier, line 68: nit: indentation?

grmartin: yeah i do quality last

JadeyOlivier (approved):
Looks good. We can smoke test full pipeline flow once merged.

PR #234, opened 2024-06-05
feat: CT refund transaction from Stripe dashboard refund

JadeyOlivier:
Updated the Stripe Webhookview to listen to the 'charge.refunded' event for if someone makes a refund directly from the Stripe dashboard. If it is a CT order, we create a refund transaction for it.

Ticket: https://2u-internal.atlassian.net/browse/SONIC-504
Functionality information ticket: https://2u-internal.atlassian.net/browse/SONIC-520

### DIFF
--- a/commerce_coordinator/apps/stripe/views.py
+++ b/commerce_coordinator/apps/stripe/views.py
@@ -12,6 +12,7 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from commerce_coordinator.apps.core.constants import PaymentState
+from commerce_coordinator.apps.rollout.utils import is_legacy_order
 from commerce_coordinator.apps.stripe.constants import StripeEventType
 from commerce_coordinator.apps.stripe.exceptions import (
     InvalidPayloadAPIError,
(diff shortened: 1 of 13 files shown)


grmartin:
The only direct comment i have here is its the first view without a filter -> pipeline... but im ok in this instance since we need to fail immediately.... but does the celery task do so? or are we trying to be faster than the webhook delay/response timout?

grmartin (approved):

commerce_coordinator/apps/stripe/views.py
```diff
@@ -26,9 +28,75 @@
 stripe.api_key = settings.PAYMENT_PROCESSOR_CONFIG['edx']['stripe']['secret_key']
 endpoint_secret = settings.PAYMENT_PROCESSOR_CONFIG['edx']['stripe']['webhook_endpoint_secret']
 source_system_identifier = settings.PAYMENT_PROCESSOR_CONFIG['edx']['stripe']['source_system_identifier']
-
-
-class WebhookView(APIView):
+NOTIFICATION_CACHE_TTL_SECS = 60 * 10  # 10 Mins
+
+
+class SingleInvocationAPIView(APIView):
```
grmartin, line 34: If we duped this can we just move the ct one in to core?

grmartin (approved):

grmartin (approved):

PR #237, opened 2024-06-11
feat: Utility script to remove GH Workflow runs

grmartin:


gh-delete-workflow-runs.sh
```diff
@@ -0,0 +1,82 @@
...
+)
+
+if [[ -z ${RUNS[*]} ]]; then
+    echo "Found no completed runs for workflow $WORKFLOW_NAME."
+    exit 0
+fi
+
+echo "Found $(echo "$RUNS" | wc -l) completed runs for workflow $WORKFLOW_NAME"
+
+for RUN in $RUNS; do
```
christopappas, line 77: [question] can you iterate over an array in bash like this?

i've had to do it differently in the past, like:
```
## now loop through the above array
for i in "${arr[@]}"
do
   echo "$i"
   # or do whatever with individual element of the array
done
```
ref: https://stackoverflow.com/questions/8880603/loop-through-an-array-of-strings-in-bash

christopappas: specifically the `"${arr[@]}"` part i'm thinking about (Instead of `$RUNS` would have assumed it would need to be `${RUNS[@]}`

grmartin: Yes its a newer bash thing, and meets the requirements of modern MacOSes (stock bash and zsh support it)

grmartin: Has to do with how the GH command outputs data i think.

christopappas: interesting okay

christopappas (approved):

PR #243, opened 2024-06-19
fix: added check for duplicate refund transactions

JadeyOlivier:
Fix following #234 and #240 which adds a check for any existing full refunds before creating a refund transaction when a `charge.refunded` stripe event fires, ensuring no duplicate refund transaction records on the Merchant Center. The signal configuration is also added to `base.py` again.
JIRA: https://2u-internal.atlassian.net/browse/SONIC-589


MushtaqRossier (approved):
Looks good to me :saluting_face:

grmartin (approved):
LGTM

PR #268, opened 2024-08-27
feat: redirect program purchases to legacy from lms/redirect

aht007:
**Merge checklist:**

**Post-merge:**

### DIFF
--- a/commerce_coordinator/apps/lms/views.py
+++ b/commerce_coordinator/apps/lms/views.py
@@ -63,7 +63,6 @@ class PaymentPageRedirectView(APIView):
             logger.exception(f"Something went wrong! Exception raised in {self.get.__name__} with error {repr(e)}")
             return HttpResponseBadRequest('Something went wrong.')
 
-
     def _redirect_response_payment(self, request):
         """
         Redirect to desired checkout view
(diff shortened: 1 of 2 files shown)


commerce_coordinator/apps/lms/views.py
```diff
@@ -74,16 +74,22 @@ def _redirect_response_payment(self, request):
...
+        if "bundle" in dict(get_items):
+            ecom_url = urljoin(
+                settings.ECOMMERCE_URL, settings.ECOMMERCE_ADD_TO_BASKET_API_PATH
+            )
+            redirect_url = self._add_query_params_to_redirect_url(ecom_url, get_items)
+        else:
+            redirect_url_obj = PaymentPageRedirectRequested.run_filter(request)
+            redirect_url = self._add_query_params_to_redirect_url(
+                redirect_url_obj["redirect_url"], get_items
+            )
```
shafqatfarhan, line 87: Is there a reason not to add this new redirect logic to the pipeline steps? cc: @grmartin mind sharing your thoughts on this approach?

shafqatfarhan (approved):

commerce_coordinator/apps/lms/views.py
aht007: The current pipeline steps are very specific to their own use case. Each pipeline step is descriptive of the work they are doing and I didn't want to add an additional responsibility in those steps.
I gave a thought to adding a new pipeline step as well but it seemed like an overkill and added complexity for a temporary measure that will be going away in near future.

grmartin: I think this is simple enough to pass for now. Ecomm is going away, itll be simpler to remove this way. Thanks for the check in though @shafqatfarhan

grmartin (approved):

PR #303, opened 2024-11-27
SONIC-704: Add CT discount availed check for outline tab

NoyanAziz:
**Merge checklist:**

**Post-merge:**


commerce_coordinator/apps/lms/views.py
```diff
@@ -334,3 +335,24 @@ def post(self, request) -> Response:
...
+        """Return True if user is eligible for a first time discount."""
+        email = request.query_params.get('email')
+
+        if not email:  # pragma: no cover
+            raise PermissionDenied(detail="Could not detect user email.")
+
+        result = CheckFirstTimeDiscountEligibility.run_filter(email=email)
+
+        output = {
+            "is_eligible": result.get('is_eligible', True)
```
syedsajjadkazmii, line 355: should the default value be True?

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -603,3 +603,36 @@ def retire_customer_anonymize_fields(self, customer_id: str, customer_version: i
...
+        try:
+            discounts = self.base_client.discount_codes.query(
+                where="code in :discountCodes",
+                predicate_var={'discountCodes': settings.COMMERCETOOLS_FIRST_TIME_DISCOUNTS}
+            )
+            discount_ids = [discount.id for discount in discounts.results]
+
+            discounted_orders = self.base_client.orders.query(
+                where=[
+                    "customerEmail=:email",
```
aht007, line 623: We can maybe save two API calls from being made here.
First thought on my mind was to use reference expansion which I couldn't find any for discounts. I think you have also checked this.
The second option here is to store the discount IDs themselves in settings/edx-internal.
We are already following this pattern for line_item_transition: [Reference](https://github.com/edx/terraform/blob/899d2e9f51d70006ac82b9e7a4bdc437dfea3a80/plans/edx/apps/commerce-coordinator/prod.tfvars#L27C35-L27C71)
It is not normally suggested to store the database identifiers like this but as long as the discounts remain the same, the ids/keys are going to be the same so we can minimize the extra API call here which makes more sense to me.
@shafqatfarhan to weigh in on this.

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -603,3 +603,36 @@ def retire_customer_anonymize_fields(self, customer_id: str, customer_version: i
...
+                    "orderState=:orderState",
+                    "discountCodes(discountCode(id in :discountIds))"
+                ],
+                predicate_var={'email': email, 'discountIds': discount_ids, 'orderState': 'Complete'}
```
aht007, line 627: Just making sure that we don't want to cater for refunded orders here right?

commerce_coordinator/apps/lms/views.py
NoyanAziz: For the default case, the user is going to be eligible for the first-time discount.

commerce_coordinator/apps/commercetools/clients.py
NoyanAziz: For the first point, we cannot query expanded objects. For the second option, I wanted the identifiers to be consistent across all environments, which is why I suggested using discount codes, as storing IDs didn’t seem ideal. However, if we're okay with this approach, we can certainly save an API call here

commerce_coordinator/apps/commercetools/clients.py
NoyanAziz: Refunded orders are not handled by CT itself, and if I remember correctly, we had to check for their first usage only. Correct me if I am wrong

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -613,3 +613,39 @@ def retire_customer_anonymize_fields(self, customer_id: str, customer_version: i
...
+                    "discountCodes(discountCode is defined)"
+                ],
+                predicate_var={'email': email, 'orderState': 'Complete'},
+                expand=["discountCodes[*].discountCode"]
+            )
+
+            if discounted_orders.total < 1:
+                return True
+
+            discounted_orders = discounted_orders.results
```
syedsajjadkazmii, line 639: shouldn't this be?
```suggestion
            if discounted_orders.results.total < 1:
                return True

            discounted_orders = discounted_orders.results
```

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -613,3 +613,39 @@ def retire_customer_anonymize_fields(self, customer_id: str, customer_version: i
...
+
+            for order in discounted_orders:
+                discount_code = order.discount_codes[0].discount_code.obj.code
```
syedsajjadkazmii, line 642: why only `[0]` index?

syedsajjadkazmii (approved):

commerce_coordinator/apps/commercetools/tests/test_clients.py
```diff
@@ -803,6 +803,89 @@ def test_update_customer_with_anonymized_fields_exception(self):
...
+                            "discountCode": {
+                                "obj": {
+                                    "code": 'another-code'
+                                }
+                            }
+                        }
+                    ]
+                }
+            ]
+        }
```
abdullahwaheed, line 826: we can make it a constant since its been repeated across tests

abdullahwaheed (approved):

commerce_coordinator/apps/commercetools/clients.py
NoyanAziz: The response of order query is in this format so it is correct
```
{
  "limit": 20,
  "offset": 0,
  "count": 1,
  "total": 1,
  "results": []
 }
```

commerce_coordinator/apps/commercetools/clients.py
NoyanAziz: The way expansion for discountCodes[*].discountCode will result in discountCodes still being an array but only having a single entry. Hence I am accessing the only value at 0th index here

commerce_coordinator/apps/commercetools/tests/test_clients.py
NoyanAziz: Both mock_orders have different code values hence, I didn't make it a constant

commerce_coordinator/settings/base.py
```diff
@@ -399,6 +399,12 @@ def root(*path_fragments):
         "pipeline": [
             'commerce_coordinator.apps.commercetools.pipeline.AnonymizeRetiredUser',
         ]
+    },
+    'org.edx.coordinator.lms.check.first.time.discount.eligibility.v1': {
+        'fail_silently': False,
+        'pipeline': [
+            'commerce_coordinator.apps.commercetools.pipeline.CheckCommercetoolsDiscountEligibility',
+        ]
```
shafqatfarhan, line 407: As discussed, we can remove this.

commerce_coordinator/apps/lms/views.py
```diff
@@ -334,3 +335,25 @@ def post(self, request) -> Response:
             logger.exception(f"[RefundView] Exception raised in {self.post.__name__} with error {repr(e)}")
             return Response('Exception occurred while retiring Commercetools customer',
                             status=HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+class FirstTimeDiscountEligibleView(APIView):
+    """View to check if a user is eligible for a first time discount"""
+    permission_classes = [LoginRedirectIfUnauthenticated]
```
shafqatfarhan, line 342:
```suggestion
    permission_classes = [IsAdminUser]
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -334,3 +335,25 @@ def post(self, request) -> Response:
...
+    throttle_classes = [UserRateThrottle]
```
shafqatfarhan, line 343:
```suggestion
    throttle_classes = (UserRateThrottle,)
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -334,3 +335,25 @@ def post(self, request) -> Response:
...
+
+    def get(self, request):
```
shafqatfarhan, line 345:
```suggestion
    def post(self, request):
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -334,3 +335,25 @@ def post(self, request) -> Response:
...
+        """Return True if user is eligible for a first time discount."""
+        email = request.query_params.get('email')
+        code = request.query_params.get('code')
```
shafqatfarhan, line 348: Please serialize these payloads

commerce_coordinator/apps/lms/views.py
```diff
@@ -334,3 +335,25 @@ def post(self, request) -> Response:
...
+
+        if not email or not code:  # pragma: no cover
+            raise PermissionDenied(detail="Could not detect user email or discount code.")
+
+        result = CheckFirstTimeDiscountEligibility.run_filter(email=email, code=code)
```
shafqatfarhan, line 353:
```suggestion
```

shafqatfarhan (approved):

PR #314, opened 2024-12-20
feat: update line items state in single api call

syedsajjadkazmii:
update line items state in single api call in fulfillment task

**Merge checklist:**

**Post-merge:**

### DIFF
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -620,7 +620,7 @@ class CommercetoolsAPIClient:
             handle_commercetools_error(err, f"Unable to update LineItemState of order {order_id}", True)
             return None
 
-    def update_all_line_items_transition_state_on_fulfillment(
+    def update_line_items_transition_state(
             self,
             order_id: str,
             order_version: int,
@@ -644,8 +644,10 @@ class CommercetoolsAPIClient:
         from_state_key = self.get_state_by_id(from_state_id).key
 
         logger.info(
-            f"[CommercetoolsAPIClient] - Transitioning all line item states for order with ID {order_id} "
-            f"from {from_state_key} to {new_state_key}"
+            f"[CommercetoolsAPIClient] - Transitioning line item states for order ID '{order_id}'. "
+            f"From State: '{from_state_key}' "
+            f"To State: '{new_state_key}' "
+            f"Line Item IDs: {', '.join(item.id for item in line_items)}"
         )
 
         try:
@@ -660,13 +662,11 @@ class CommercetoolsAPIClient:
                     for item in line_items
                 ]
 
-                updated_fulfillment_line_item_order = self.base_client.orders.update_by_id(
+                return self.base_client.orders.update_by_id(
                     id=order_id,
                     version=order_version,
                     actions=actions,
                 )
-
-                return updated_fulfillment_line_item_order
             else:
                 logger.info(
                     f"All line items already have the correct state {new_state_key}. "
@@ -675,7 +675,12 @@ class CommercetoolsAPIClient:
                 return self.get_order_by_id(order_id)
         except CommercetoolsError as err:
             # Logs & ignores version conflict errors due to duplicate Commercetools messages
-            handle_commercetools_error(err, f"Unable to update all LineItemStates of order {order_id}", True)
+            handle_commercetools_error(
+                err,
+                f"Failed to update LineItemStates for order ID '{order_id}'. "
+                f"Line Item IDs: {', '.join(item.id for item in line_items)}",
+                True
+            )
             return None
 
     def retire_customer_anonymize_fields(
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -16,6 +16,7 @@ from commercetools.platform.models import CustomerChangeEmailAction, CustomerSet
 from commercetools.platform.models import CustomerSetCustomTypeAction as CTCustomerSetCustomTypeAction
 from commercetools.platform.models import CustomerSetFirstNameAction, CustomerSetLastNameAction
 from commercetools.platform.models import FieldContainer as CTFieldContainer
+from commercetools.platform.models import LineItem as CTLineItem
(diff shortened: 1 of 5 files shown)


commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -619,6 +620,64 @@ def update_line_item_transition_state_on_fulfillment(
...
+                    for item in line_items
+                ]
+
+                updated_fulfillment_line_item_order = self.base_client.orders.update_by_id(
+                    id=order_id,
+                    version=order_version,
+                    actions=actions,
+                )
+
+                return updated_fulfillment_line_item_order
```
abdullahwaheed, line 669: nit: 
```suggestion
                return self.base_client.orders.update_by_id(
                    id=order_id,
                    version=order_version,
                    actions=actions,
                )
```

abdullahwaheed (approved):

NawfalAhmed (approved):

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -619,6 +620,64 @@ def update_line_item_transition_state_on_fulfillment(
             handle_commercetools_error(err, f"Unable to update LineItemState of order {order_id}", True)
             return None
 
+    def update_all_line_items_transition_state_on_fulfillment(
```
aht007, line 623: Can we rename this to just `update_line_items_transition_state`.
Two reasons:
We are passing a list in params so it is automatically inferred that it will cater all those hence we don't need the update_all thing.
Secondly we can use this utility in generic sense and not just for updating the state due to fulfillment only hence why we need to skip on_fulfillment part and just give the sense that the utility will transition from from_state_id to new_state_key

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -619,6 +620,64 @@ def update_line_item_transition_state_on_fulfillment(
...
+            new_state_key (str): Key of LineItemState to transition to
+        Returns (CTOrder): Updated order object or
+        Returns (CTOrder): Current un-updated order
+        Raises Exception: Error if update was unsuccessful.
+        """
+
+        from_state_key = self.get_state_by_id(from_state_id).key
+
+        logger.info(
+            f"[CommercetoolsAPIClient] - Transitioning all line item states for order with ID {order_id} "
```
aht007, line 647: Can we move this to a line below and also add lineItem ids that we are sending in the API call for reference and ease of debugging.

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -619,6 +620,64 @@ def update_line_item_transition_state_on_fulfillment(
...
+
+                updated_fulfillment_line_item_order = self.base_client.orders.update_by_id(
+                    id=order_id,
+                    version=order_version,
+                    actions=actions,
+                )
+
+                return updated_fulfillment_line_item_order
+            else:
+                logger.info(
```
aht007, line 671: Is there a chance that some of our lineItems got updated but some of them didn't? What would be the flow in that case?

commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -619,6 +620,64 @@ def update_line_item_transition_state_on_fulfillment(
...
+                    f"All line items already have the correct state {new_state_key}. "
+                    "Not attempting to transition LineItemState"
+                )
+                return self.get_order_by_id(order_id)
+        except CommercetoolsError as err:
+            # Logs & ignores version conflict errors due to duplicate Commercetools messages
+            handle_commercetools_error(err, f"Unable to update all LineItemStates of order {order_id}", True)
```
aht007, line 678: nit: Maybe add lineItems here as well to the logging and remove the word `all `from text as it looks redundant.

commerce_coordinator/apps/commercetools/sub_messages/tasks.py
```diff
@@ -98,19 +98,22 @@ def fulfill_order_placed_message_signal_task(
     canvas_entry_properties = {"products": []}
     canvas_entry_properties.update(extract_ct_order_information_for_braze_canvas(customer, order))
 
+    logger.info(
```
aht007, line 101: I am not sure if we should do it this way since: 
This task is triggered from the fulfill endpoint which is called when the lineItem state of an item in the order is updated to fulfillment pending state.
What that means is we would be triggering a number of duplicate calls for fulfilling all the lineItems.
We might need to change our workflow and think about this perspective

aht007 (commented):
Left a few comments.

commerce_coordinator/apps/commercetools/sub_messages/tasks.py
aht007: This thing is being catered in the [view](https://github.com/edx/commerce-coordinator/blob/59ab5f85d39f063a32017cc7dd9b52bd2f4fcb67/commerce_coordinator/apps/commercetools/views.py#L53) itself where we are marking the task as running based on `order_id`. Hence we don't need to worry about it and I am marking it as resolved.

commerce_coordinator/apps/commercetools/clients.py
syedsajjadkazmii: nope. its transactional.

https://docs.commercetools.com/api/performance-tips#multiple-update-actions

commerce_coordinator/apps/commercetools/clients.py
syedsajjadkazmii: done

commerce_coordinator/apps/commercetools/clients.py
syedsajjadkazmii: done

commerce_coordinator/apps/commercetools/clients.py
syedsajjadkazmii: done

commerce_coordinator/apps/commercetools/clients.py
aht007: Previously I was of the view that we were fetching lineitems in the below line but I have gone through it again and they are available at this point. Resolving this comment.

aht007 (approved):
LGTM

PR #337, opened 2025-02-14
feat: Part of Program refund in Commerce Coordinator

abdullahwaheed:
Ticket
[SONIC-866](https://2u-internal.atlassian.net/browse/SONIC-866)

### DIFF
--- a/commerce_coordinator/apps/lms/utils.py
+++ b/commerce_coordinator/apps/lms/utils.py
@@ -3,9 +3,16 @@
 from commerce_coordinator.apps.commercetools.clients import CommercetoolsAPIClient
 
 
-def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> str:
+def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> tuple[str, str]:
     """
-    Check if line item with given entitlement is present.
+    Retrieve the order ID and line item ID associated with the given entitlement ID.
+
+    Args:
+        order_number (str): The order number in Commercetools.
+        entitlement_id (str): The entitlement ID to search for.
+
+    Returns:
+        tuple[str, str]: A tuple containing the order ID and the matching line item ID (if found, otherwise an empty string).
     """
     ct_api_client = CommercetoolsAPIClient()
     ct_order = ct_api_client.get_order_by_number(order_number=order_number)
@@ -15,5 +22,6 @@ def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> st
     for line_item in ct_order.line_items:
         if line_item.custom.fields.get('edxLMSEntitlementId', '') == entitlement_id:
             order_line_item_id = line_item.id
+            break
 
     return order_id, order_line_item_id
--- a/commerce_coordinator/apps/lms/utils.py
+++ b/commerce_coordinator/apps/lms/utils.py
@@ -1,0 +1,19 @@
+"""LMS Utility Functions"""
+
+from commerce_coordinator.apps.commercetools.clients import CommercetoolsAPIClient
+
+
+def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> str:
+    """
+    Check if line item with given entitlement is present.
+    """
+    ct_api_client = CommercetoolsAPIClient()
+    ct_order = ct_api_client.get_order_by_number(order_number=order_number)
+
+    order_id = ct_order.id
+    order_line_item_id = ''
+    for line_item in ct_order.line_items:
+        if line_item.custom.fields.get('edxLMSEntitlementId', '') == entitlement_id:
+            order_line_item_id = line_item.id
+
+    return order_id, order_line_item_id
--- a/commerce_coordinator/apps/lms/utils.py
+++ b/commerce_coordinator/apps/lms/utils.py
@@ -12,7 +12,8 @@ def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> tu
         entitlement_id (str): The entitlement ID to search for.
 
     Returns:
-        tuple[str, str]: A tuple containing the order ID and the matching line item ID (if found, otherwise an empty string).
+        tuple[str, str]: A tuple containing the order ID and the matching line item ID (if found,
+        otherwise an empty string).
     """
     ct_api_client = CommercetoolsAPIClient()
     ct_order = ct_api_client.get_order_by_number(order_number=order_number)
--- a/commerce_coordinator/apps/lms/views.py
+++ b/commerce_coordinator/apps/lms/views.py
@@ -31,6 +31,7 @@ from commerce_coordinator.apps.lms.serializers import (
     UserRetiredInputSerializer,
     enrollment_attribute_key
 )
+from commerce_coordinator.apps.lms.utils import get_line_item_from_entitlement
 from commerce_coordinator.apps.rollout.utils import is_legacy_order
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


commerce_coordinator/apps/lms/views.py
```diff
@@ -247,22 +273,31 @@ def post(self, request) -> Response:
...
+            ) if entitlement_refund else (
+                f"[RefundView] Failed processing refund for order {order_id} for username: {username}, "
+                f"course_id: {course_id} the enrollment_attributes array requires an orders: line_item_id "
+                f"attribute."
+            )
+            logger.error(log_msg)
+
+            response_msg = 'Line item does not exist.' if entitlement_refund else \
+                'the enrollment_attributes array requires an orders: line_item_id attribute.'
+            return Response(response_msg, status=HTTP_400_BAD_REQUEST)
```
syedsajjadkazmii, line 288: NIT
```suggestion
if not order_line_item_id:
    log_msg_prefix = f"[RefundView] Failed processing refund for order {order_id} for username: {username}, "
    
    if entitlement_refund:
        log_msg = f"{log_msg_prefix}entitlement_id: {entitlement_id}. Line item does not exist."
        response_msg = "Line item does not exist."
    else:
        log_msg = (
            f"{log_msg_prefix}course_id: {course_id}. The enrollment_attributes array "
            "requires an orders: line_item_id attribute."
        )
        response_msg = "The enrollment_attributes array requires an orders: line_item_id attribute."

    logger.error(log_msg)
    return Response(response_msg, status=HTTP_400_BAD_REQUEST)
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -247,22 +273,31 @@ def post(self, request) -> Response:
...
 
         if not order_line_item_id:
-            logger.error(f"[RefundView] Failed processing refund for order {order_id} for username: {username}, "
-                         f"course_id: {course_id} the enrollment_attributes array requires an orders: line_item_id "
-                         f"attribute.")
-            return Response('the enrollment_attributes array requires an orders: line_item_id '
-                            'attribute.', status=HTTP_400_BAD_REQUEST)
+            log_msg = (
+                f"[RefundView] Failed processing refund for order {order_id} for username: {username}, "
+                f"entitlement_id: {entitlement_id}. Line item does not exist."
```
syedsajjadkazmii, line 278:
```suggestion
                f"entitlement_id: {entitlement_id}. Line item does not exist on commercetools Order."
```

commerce_coordinator/apps/lms/views.py
```diff
@@ -219,25 +221,49 @@ def post(self, request) -> Response:
...
 
-        logger.info(f"[RefundView] Starting LMS Refund for username: {username}, course_id: {course_id}, "
-                    f"Enrollment attributes: {enrollment_attributes}.")
+            logger.info(f"[RefundView] Starting LMS Refund for username: {username}, course_id: {course_id}, "
+                        f"Enrollment attributes: {enrollment_attributes}.")
 
-        order_line_item_id = enrollment_attributes.get(enrollment_attribute_key('order', 'line_item_id'), None)
-        order_id = enrollment_attributes.get(enrollment_attribute_key('order', 'order_id'), None)
+            order_line_item_id = enrollment_attributes.get(enrollment_attribute_key('order', 'line_item_id'), None)
+            order_id = enrollment_attributes.get(enrollment_attribute_key('order', 'order_id'), None)
```
syedsajjadkazmii, line 266: the order_id verification should be inside the else block right? because `if not order_id` cannot be the case for entitlements refund? if there is not order id, it should be handled as part of serializer validation, right?

but if we still want to put `if not order_id` outside of if-else, we should make the message generic.

commerce_coordinator/apps/lms/views.py
```diff
@@ -247,22 +273,31 @@ def post(self, request) -> Response:
...
+            logger.error(log_msg)
+
+            response_msg = 'Line item does not exist.' if entitlement_refund else \
+                'the enrollment_attributes array requires an orders: line_item_id attribute.'
+            return Response(response_msg, status=HTTP_400_BAD_REQUEST)
 
         try:
             result = OrderRefundRequested.run_filter(order_id, order_line_item_id)
 
+            log_type = f"{entitlement_id=}" if entitlement_refund else f"{course_id=}"
```
syedsajjadkazmii, line 293:
```suggestion
            info_log_details = f"{entitlement_id=}" if entitlement_refund else f"{course_id=}"
```

commerce_coordinator/apps/lms/utils.py
syedsajjadkazmii, line 1: please write unit tests for this file.

commerce_coordinator/apps/lms/views.py
syedsajjadkazmii, line 1: please update/write unit tests

commerce_coordinator/apps/lms/utils.py
```diff
@@ -0,0 +1,19 @@
+"""LMS Utility Functions"""
+
+from commerce_coordinator.apps.commercetools.clients import CommercetoolsAPIClient
+
+
+def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> str:
```
syedsajjadkazmii, line 6:
```suggestion
def get_line_item_from_entitlement(order_number: str, entitlement_id: str) -> tuple[str, str]:
```

commerce_coordinator/apps/lms/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Check if line item with given entitlement is present.
+    """
```
syedsajjadkazmii, line 9:
```suggestion
  """
    Retrieve the order ID and line item ID associated with the given entitlement ID.

    Args:
        order_number (str): The order number in Commercetools.
        entitlement_id (str): The entitlement ID to search for.

    Returns:
        tuple[str, str]: A tuple containing the order ID and the matching line item ID (if found, otherwise an empty string).
    """
```

commerce_coordinator/apps/lms/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    ct_api_client = CommercetoolsAPIClient()
+    ct_order = ct_api_client.get_order_by_number(order_number=order_number)
+
+    order_id = ct_order.id
+    order_line_item_id = ''
+    for line_item in ct_order.line_items:
+        if line_item.custom.fields.get('edxLMSEntitlementId', '') == entitlement_id:
+            order_line_item_id = line_item.id
```
syedsajjadkazmii, line 17:
```suggestion
    order_line_item_id = next(
        (line_item.id for line_item in ct_order.line_items if line_item.custom.fields.get('edxLMSEntitlementId') == entitlement_id),
        ''
    )
```

Instead of looping manually and assigning a variable, next() finds the first match and returns immediately.

commerce_coordinator/apps/lms/views.py
abdullahwaheed: if in future we need to update the whole log together for entitlement, we have to update the prefix. Moreover, inline conditioning is better for such cases

commerce_coordinator/apps/lms/views.py
abdullahwaheed: serializer is expecting `order_number` for entitlement so order id won't be verified in serializer. If an order does not found against the given order number, an exception is raised and returned. So in case, order id cannot be empty. Regarding adding it in else, yes we can safely move it to course code

commerce_coordinator/apps/lms/views.py
abdullahwaheed: its log type, not log detail

syedsajjadkazmii: umm, i thought we are just storing what data to print in log in `log_type` and log type refers to types such as `warn` `debug` and `info` logs.

syedsajjadkazmii (approved):

PR #360, opened 2025-03-10
fix: fixed InvalidFilterType exception, Missing Transaction ID and Incorrect Title in refund flow

abdullahwaheed:
Ticket
[Handle InvalidFilterType exception, Missing Transaction ID and Incorrect Title in refund flow](https://2u-internal.atlassian.net/browse/SONIC-936)

### DIFF
--- a/commerce_coordinator/apps/stripe/pipeline.py
+++ b/commerce_coordinator/apps/stripe/pipeline.py
@@ -189,4 +189,10 @@ class RefundPaymentIntent(PipelineStep):
             logger.info(f'[CT-{tag}] Unsuccessful Stripe refund with details:'
                         f'[order_id: {order_id}, payment_intent_id: {payment_intent_id}'
                         f'message_id: {kwargs["message_id"]}')
+
+            if 'greater than unrefunded amount on charge' in ex.user_message:
+                return {
+                    'failed_psp': ex.user_message
+                }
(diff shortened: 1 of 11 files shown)


commerce_coordinator/apps/stripe/pipeline.py
```diff
@@ -189,4 +189,10 @@ def run_filter(
             logger.info(f'[CT-{tag}] Unsuccessful Stripe refund with details:'
                         f'[order_id: {order_id}, payment_intent_id: {payment_intent_id}'
                         f'message_id: {kwargs["message_id"]}')
+
+            if 'greater than unrefunded amount on charge' in ex.user_message:
```
abdullahwaheed, line 193: we didn't have any other param except this message to identify this. `ex.code` is empty, `ex.param` is `amount`, which is not reliable.

commerce_coordinator/apps/stripe/pipeline.py
```diff
@@ -189,4 +189,10 @@ def run_filter(
...
+                return {
+                    'failed_psp': ex.user_message
```
shafqatfarhan, line 195:
```suggestion
                    'psp_refund_error': ex.user_message
```

shafqatfarhan (approved):

PR #371, opened 2025-03-20
fix: fix log stream issue locally

mubbsharanwar:
We are moving the logging format string to the remote config to fix the logging issue locally.

We can set LOGGING_FORMAT_STRING in our private.py file to get logs for local env like:

`LOGGING_FORMAT_STRING = '%(asctime)s %(levelname)s %(process)d [%(name)s] [user %(userid)s] [ip %(remoteip)s] %(filename)s:%(lineno)d - %(message)s`'

[edx-internal](https://github.com/edx/edx-internal/pull/12516)

**Merge checklist:**

**Post-merge:**

### DIFF
--- a/commerce_coordinator/settings/local.py
+++ b/commerce_coordinator/settings/local.py
@@ -113,7 +113,8 @@ JWT_AUTH.update({
 
 ENABLE_AUTO_AUTH = True
 
-LOGGING = get_logger_config(debug=DEBUG)
+LOGGING_FORMAT_STRING = os.environ.get("LOGGING_FORMAT_STRING", "")
+LOGGING = get_logger_config(debug=DEBUG, format_string=LOGGING_FORMAT_STRING)
 
 CELERY_BROKER_URL = "redis://:password@localhost:6379/0"
 
--- a/commerce_coordinator/settings/local.py
+++ b/commerce_coordinator/settings/local.py
@@ -113,8 +113,7 @@ JWT_AUTH.update({
 
 ENABLE_AUTO_AUTH = True
 
-LOGGING_FORMAT_STRING = os.environ.get("LOGGING_FORMAT_STRING", "")
-LOGGING = get_logger_config(debug=DEBUG, format_string=LOGGING_FORMAT_STRING)
+LOGGING = get_logger_config(debug=DEBUG)
 
 CELERY_BROKER_URL = "redis://:password@localhost:6379/0"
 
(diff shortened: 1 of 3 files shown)


commerce_coordinator/settings/local.py
```diff
@@ -113,7 +113,8 @@
 
 ENABLE_AUTO_AUTH = True
 
-LOGGING = get_logger_config(debug=DEBUG)
+LOGGING_FORMAT_STRING = os.environ.get("LOGGING_FORMAT_STRING", "")
```
shafqatfarhan, line 116: This will always result in an empty string as we have not set its value for local setup anywhere. So, no need to get its value and pass it to the logger function as it already defaults to an empty string, i.e., `format_string=''`

commerce_coordinator/settings/utils.py
```diff
@@ -39,8 +40,7 @@ def get_logger_config(logging_env="no_env",
         'disable_existing_loggers': False,
         'formatters': {
             'standard': {
-                'format': '%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] '
-                          '[dd.trace_id=%(dd.trace_id)s dd.span_id=%(dd.span_id)s] - %(message)s',
+                'format': format_string,
```
shafqatfarhan, line 43: For local logging, this change will result in an empty format string. Previously, it had a default format that was used for all the environments, but with this change, I am not sure what format the local setup will follow.

commerce_coordinator/settings/local.py
mubbsharanwar: In this way, we can set its value in private.py. No?

commerce_coordinator/settings/utils.py
mubbsharanwar: We can set the default format in private.py that's why I am getting it in settings/local.py using `environ`

mubbsharanwar: We set the previously used formate as a default for local setup.

commerce_coordinator/settings/production.py
```diff
@@ -10,7 +10,8 @@
 
 ALLOWED_HOSTS = ['*']
 
-LOGGING = get_logger_config()
+LOGGING_FORMAT_STRING = os.environ.get("LOGGING_FORMAT_STRING", "")
+LOGGING = get_logger_config(format_string=LOGGING_FORMAT_STRING)
```
shafqatfarhan, line 14:
```suggestion
LOGGING = get_logger_config(format_string=LOGGING_FORMAT_STRING) if LOGGING_FORMAT_STRING else get_logger_config()
```

shafqatfarhan (approved):

commerce_coordinator/settings/utils.py
abdullahwaheed: the default would work for tests as well

commerce_coordinator/settings/local.py
abdullahwaheed: this change has been reverted. now we rely on `DEFAULT_LOGGING_FORMAT_STRING`

PR #388, opened 2025-04-21
feat: ADR for fulfillment webhook implementation solutions

syedsajjadkazmii:
ADR for Webhook communication solutions between commerce-coordinator and order-fulfillment service.

**Merge checklist:**

**Post-merge:**


shafqatfarhan (approved):

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+We are designing a new fulfillment engine as a Django-based service (Order Fulfillment, or OF) to process fulfillment (e.g., fulfilling in LMS) requests initiated by the Commerce Coordinator (CC) service. These services will be maintained by different organizations — OF by 2U, and CC by Red Ventures.
+
+To maintain clear service boundaries and reduce coupling, we need a secure, reliable, and decoupled communication model. The mechanism must work across platforms, be framework-agnostic, and ensure message integrity and authenticity — similar to webhooks used by platforms like Stripe and PayPal.
+
+This ADR documents the different implementation approaches and rationale for choosing a custom webhook-based approach.
+
+Decision: Custom Webhook-Based Implementation
+*********************************************
+
+We will implement a custom, HMAC-secured webhook mechanism, inspired by Stripe, for communication between services. The key security feature of this approach is the use of a shared secret key between the Commerce Coordinator (CC) and Order Fulfillment (OF) services to generate and verify HMAC (Hash-based Message Authentication Code) signatures.
```
syedsajjadkazmii, line 22: Hey @colinbrash — this authentication approach (HMAC with timestamp) is widely used and adopted by providers like Stripe. Let us know if we should still get it validated by the Cybersecurity team.

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Status
+******
+
+**Draft**
+
+Context
+*******
+
+We are designing a new fulfillment engine as a Django-based service (Order Fulfillment, or OF) to process fulfillment (e.g., fulfilling in LMS) requests initiated by the Commerce Coordinator (CC) service. These services will be maintained by different organizations — OF by 2U, and CC by Red Ventures.
```
christopappas, line 13: As I understand it today, CC is not going to be maintained by Red Ventures. Any functionality from "commerce-coordinator-esque" that RV will want to maintain will be recreated on the RV side.

This is due to operational reasons: RV employees cannot gain VPN access presently, and so cannot deploy or troubleshoot commerce-coordinator in depth, as well as RV employees cannot (should not) run devstack, as there is far too much edx specific context required to run this tool when most of the work we'll be doing is no longer in that ecosystem.

The future is looking like some endpoint will exist on 2U's side (probably the fulfillment service), and something on the RV side will call it after we accept payment upstream.

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+- **Django-Specific**: Primarily designed for Django-based services, which is limiting since future consumers may not use Django.
+
+The strong dependency on Open edX and Django conflicts with our architectural goals of service independence and platform neutrality. Thus, we ruled this option out.
+
+**References**
+
+- `How to start using the Event Bus (Open edX) <https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus>`_
+- `How to Use the Event Bus on edX.org (2U) <https://2u-internal.atlassian.net/wiki/spaces/AT/pages/174555142/How+to+Use+the+Event+Bus+edX.org+2+of+2>`_
+
+2. AWS EventBridge
```
christopappas, line 171: The team at RV is actually planning to use EventBridge with CommerceTools... I think it may make sense for us to chat through some of this (myself, Colin, Glenn, and Juliana). Want to make sure we're aligned on the approach 🙏

christopappas (changes requested):
Re-reviewing with "request changes" because I think we should have a discussion before we move ahead with this approach. Thanks!

docs/decisions/0009-webhook-based-communication.rst
colinbrash: I don't think adding addition information about the intended future state of CC is needed for this ADR.

This ADR covers only the splitting out of the OF service, and I think it accurately states that once the OF service is split out, ownership will transfer to 2U, and CC (as long as it continues to exist) will continue to be maintained by RV.

docs/decisions/0009-webhook-based-communication.rst
colinbrash: We don't need this to block our work, but we should create a Confluence page specifically for the auth approach and point CyberSec to it, so they have an opportunity to ask questions or voice concerns.

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+- **Encrypted Shared Key**: They shared key will be encrypted and stored in `edx-internal` for both CC and OF.
+- **Proven Pattern**: Inspired by widely-used, industry-standard webhook models (e.g., Stripe), which are well-tested and understood.
+- **Cross-Platform Compatible**: Works across services implemented in different tech stacks, with no dependency on platform-specific features or SDKs.
+- **Decoupled Architecture**: Clean separation between services thus reducing service coupling.
+- **Flexibility Control**: Full ownership over how requests are validated, retried, and logged.
+- **No External Dependency**: Does not rely on third-party cloud infrastructure (e.g., AWS EventBridge), enabling more flexibility and control.
+- **Zero Infrastructure Cost**: No additional cost associated with using a cloud event bus or message queue. Relies on HTTPS and standard cryptographic libraries.
+- **Lightweight and Fast**: Low overhead in both message size and processing latency. Uses minimal resources and fast cryptographic operations.
+- **Custom Retry Logic**: Since its a custom HTTP based solution, we have the flexibility to implement retry mechanism.
```
colinbrash, line 137: I don't know what the benefit is here.  Do we _want_ custom retry logic for some reason?

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+- **No Payload Limits**: Unlike EventBridge (which enforces a ~256 KB size limit), this model allows payload sizes as needed.
```
colinbrash, line 138: I don't see the benefit here either.  In what circumstances would a fulfillment service ever need more than a 256K payload?

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+- **Single Sender–Receiver**: The communication is between one sender and one receiver and be enhanced as needed.
+
+Cons
+====
+
+- **Requires Custom Implementation**: Unlike managed services like AWS EventBridge, this solution requires us to build, test, and maintain the retry logic.
+- **Manual Key Management**: Secrets must be managed and rotated manually or via internal tooling, which adds operational overhead and potential for misconfiguration if not handled properly.
+- **No Built-in Delivery Guarantees**: Unlike EventBridge, which guarantees at-least-once delivery with retries, we need to implement our own retry mechanism.
```
colinbrash, line 146: These seem like very big cons to me.  Particularly concerning is the need to implement and maintain a custom retry mechanism.

If you feel this is not that big of a concern, could you describe more about the risks and mitigations?

docs/decisions/0009-webhook-based-communication.rst
colinbrash: I agree with Chris here.  I'd like a little more discussion about EventBridge, since it is being used elsewhere in the system already.

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+
+2. AWS EventBridge
+==================
+
+AWS EventBridge supports asynchronous service communication through an event bus and API destinations, offering a robust and managed pub/sub system.
+
+**Cons**
+
+- **Operational Cost**: Additional cost for message processing and API destinations.
+- **Infrastructure Complexity**: Setting up EventBridge involves managing IAM users, access credentials, Secrets Manager, API destinations, and event routing rules — increasing operational burden.
```
colinbrash, line 179: Could you elaborate on what the complexity would be specifically for this application?

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+- `How to Use the Event Bus on edX.org (2U) <https://2u-internal.atlassian.net/wiki/spaces/AT/pages/174555142/How+to+Use+the+Event+Bus+edX.org+2+of+2>`_
+
+2. AWS EventBridge
+==================
+
+AWS EventBridge supports asynchronous service communication through an event bus and API destinations, offering a robust and managed pub/sub system.
+
+**Cons**
+
+- **Operational Cost**: Additional cost for message processing and API destinations.
```
colinbrash, line 178: Do you have any data on what the cost would be?

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+- **Infrastructure Complexity**: Setting up EventBridge involves managing IAM users, access credentials, Secrets Manager, API destinations, and event routing rules — increasing operational burden.
+- **Local Development Overhead**: Requires mocking or local setup tools to simulate EventBridge.
+
+While AWS EventBridge provides powerful features such as built-in retries, scalable pub/sub architecture, cross-platform compatibility, and secure bidirectional messaging, the added cost and infrastructure overhead outweigh the benefits for a relatively simple point-to-point communication use case like ours.
```
colinbrash, line 182: If we were to use EventBridge, I would imagine that the operational overhead and costs would be covered by the owner of the _client_.  Meaning, OF would simply know how to publish events to the desired EB instance, much like Commercetools currently allows an integration with EB.  Is this what you are describing as well, or do we have different pictures of an EB implementation?

I would really like to understand better what the cost of implementation would be if we OF were to support an EB integration.  One benefit I see that is not mentioned would be potentially lower cost to maintain for the owner of the OF service.

docs/decisions/0009-webhook-based-communication.rst
```diff
@@ -0,0 +1,197 @@
...
+
+         hmac.compare_digest(received_signature, expected_signature)
+
+If the timestamp is valid and the signatures match, the request is processed.
+
+Pros
+====
+
+- **Secure Communication**: Since HMAC-based webhooks are a known and trusted model, there is less need for a new security review from scratch.
+- **Encrypted Shared Key**: They shared key will be encrypted and stored in `edx-internal` for both CC and OF.
```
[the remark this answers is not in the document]
syedsajjadkazmii: Yes, you're right — the intention behind that point was to highlight the security benefit of the webhook-based structure, rather than the specific use of edx-internal. The key idea is that communication is protected by a shared secret, and any request without it will be rejected.

That said, you're absolutely correct — once the client (CC) is fully within RV infrastructure, storing the key in edx-internal would no longer make sense. If we proceed with this approach, the shared secret can instead be securely stored within RV's infrastructure.

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: Good question — to clarify, retry mechanisms are important to ensure reliable delivery of the API call. Failures can happen for various reasons — server downtime, deployments in progress, or other unforeseen issues — and without retries, we risk dropping important fulfillment messages.

That said, the benefit here isn't just that we need retry logic, but that with a custom implementation, we have full flexibility in how we implement it — including conditional retries, and logging or alerting around failures.

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: You're right — this isn’t a meaningful benefit in our case. I’ll go ahead and remove this point.

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: You're right — the lack of built-in delivery guarantees is a significant downside, especially compared to EventBridge’s at-least-once delivery. However, implementing our own retry mechanism gives us the flexibility to tailor retries to our needs.

To mitigate risks such as server downtime, we can:

- Implement monitoring and alerting for failed attempts.
- Make the retry strategy configurable to adapt as needed.
- Add detailed logging for better traceability.

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: For this application, setting up EventBridge would involve creating a custom event bus and configuring two rules: one for communication from the Commerce Coordinator (CC) to the Order Fulfillment (OF) service (for the fulfillment request) and another for the reverse communication from OF to CC (to receive the fulfillment response and update the order object).

We would need to configure API destinations to define where the events are sent. IAM roles are necessary to securely grant access to the event bus, and we will need to configure access keys (access ID and secret) to allow the CC service to emit events to the custom event bus. The reason we need to create a new custom event bus is that the source in the current setup is a partner (Commercetools), whereas in this case, the source would be our own service (CC).

By infra complexity, we are referring to all of this.

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: I'm not entirely sure how the costs for AWS EventBridge will be managed or who will be responsible for that. However, I can explain the setup that would be required for this split.

We would need two-way communication between CC and OF. First, CC will send an order fulfillment request to OF, and once fulfillment is complete in LMS, OF will respond back asynchronously to CC about the success or failure of the fulfillment. This allows CC to update the order state.

To achieve this, we would create a custom event bus with two rules and two API destinations:

One API destination will be for sending the fulfillment request from CC to OF.

The other API destination will handle the fulfillment response from OF back to CC.

Both services (CC and OF) will need AWS access keys and secrets through IAM roles to allow them to publish events to the EventBridge bus.

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: I asked SRE about it, and they shared this: https://aws.amazon.com/eventbridge/pricing/

Although I am no expert in costing and analysis 😅  but giving a look at the document it seems like the cost is minimal

- Maintaining two API destination connections would cost about $1/month.
- Publishing events would be free for the first 100k events per month, and then $1 per million events beyond that.

Although the main concern for not proposing AWS was related to costs, but since the costs is minimal. we can also go with AWS as a potential solution. 

Could you please provide your opinion on this?

docs/decisions/0009-webhook-based-communication.rst
syedsajjadkazmii: just referencing: https://github.com/edx/commerce-coordinator/pull/388#discussion_r2063138785

syedsajjadkazmii: > One benefit I see that is not mentioned would be potentially lower cost to maintain for the owner of the OF service.

Considering the maintenance cost of the OF service, we don't anticipate any ongoing maintenance burden with the custom implementation beyond handling unforeseen errors via logging and retries. That said, we’ve done a POC for both the custom solution and AWS EventBridge, and we’re comfortable proceeding with either — we'd appreciate your thoughts on which direction makes more sense.

Also, with AWS EventBridge, we’d need SRE involvement to provision the event bus and set up IAM roles in both staging and production. In comparison, the custom implementation only requires SRE support for generating and managing the shared secret key.

docs/decisions/0009-webhook-based-communication.rst
colinbrash: Could we design this similar to how CT works?  We use the CT API to communicate _to_ CT.  CT uses EB to publish events.  Similarly, CC could use an OF API to request fulfillment, and OF could use EB to notify subscribers (CC) async when fulfillment is completed.

This eliminates one half of the EB need.

I also would like to know whether we can use the existing EB setup that CC already uses?

docs/decisions/0009-webhook-based-communication.rst
colinbrash: > implementing our own retry mechanism gives us the flexibility to tailor retries to our needs

This is what I question the benefit of.  What needs do we actually have that make us prefer custom retry logic over EB?

I also prefer EB because it put the bulk of the monitoring on the _consumer_.  For example, once CC logic lives at RV, if RV EB is used then the OF service only needs to monitor up to the point where it successfully publishes an event.

docs/decisions/0009-webhook-based-communication.rst
colinbrash: See my other comments -- I would prefer we go with an endpoint for the fulifillment request, and EB for publishing fulfillment completion.  I think it makes sense to use the EB setup we already have for CC?

docs/decisions/0009-webhook-based-communication.rst
colinbrash: > Making an API request from CC to OF and then from OF to LMS for fulfillment would introduce an additional synchronous call layer in the fulfillment flow. This creates similar challenges we’re trying to avoid with the custom implementation, such as handling retries and failures, which could complicate the process.

I'm not understanding the concern.  Making a synchronous call to an API is not something that should require abnormal, custom retry and failure handling.  My only concern with the webhook approach is that this does seem to require abnormal, custom retry/failure handling that EB already provides out of the box.

An event bus is used to let a publisher tell a subscriber know _what happened_, not to tell it _what to do_.  I think the OF client ought to be telling OF what to do rather than telling it what happened.  The main reason for this is because OF is not necessarily going to have a 1-1 relationship with CC.  The intention is to have a single service that is capable of fulfilling any 2U order.  In the future, OF may be used by Get Smarter, for example.  So if OF fulfills orders based on some knowledge of CC events, it will need to have additional knowledge of other systems as well, in order to fulfill orders for those systems too.

Next, when OF is done doing what is asked, then it should be telling the client what happened, and _not_ what to do.  There _is_ a 1-1 relationship between CC (or whatever the future backend service is) and OF.

Also, the work OF is doing requires an async response, while the instructions CC need to give OF do not require an async request.

Finally, providing no API at all to request a fulfillment does not make sense to me for a fulfillment service.  Doing the work and only publishing that it happened on EB feels ok -- the service can still work without the client knowing whether the request was successful.  But having no way to directly ask the service to perform a fulfillment feels like an incomplete/bad API.

So the best architectural fit for this, in my view, is to have a synchronous API for OF to be told what to do (by CC in this case), and then have OF use an event bus to async notify a subscriber (again, CC in this case) what the result of the instruction was.

If I'm off base on my analysis, let me know.  Ultimately, I am ok whichever implementation you pick.  If what I am saying makes sense, let's make a synchronous API.  If you still think EB is appropriate in both directions, go for it! :)

syedsajjadkazmii: @colinbrash, Thanks again for the thoughtful input — your analysis makes complete sense. We agree with your recommendation and will proceed with implementing a synchronous API in OF for CC to initiate fulfillment requests, and use EventBridge for OF to send asynchronous fulfillment status updates back to CC.

We’ll update the ADR to reflect this revised architecture.

Appreciate your guidance!

christopappas (approved):

PR #411, opened 2025-05-16
feat: gracefully handle mobile order refund in CC

syedsajjadkazmii:
For mobile orders, we want to simply create a return and mark it refunded without doing any psp refund.

**Merge checklist:**

**Post-merge:**

### DIFF
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -83,7 +83,7 @@ from commerce_coordinator.apps.commercetools.utils import (
     find_refund_transaction,
     get_refund_transaction_id_from_mobile_order,
     handle_commercetools_error,
-    translate_refund_status_to_transaction_status,
+    translate_refund_status_to_transaction_status
 )
 from commerce_coordinator.apps.core.constants import ORDER_HISTORY_PER_SYSTEM_REQ_LIMIT
 
@@ -899,8 +899,7 @@ class CommercetoolsAPIClient:
                 f"Line Item IDs: {', '.join(item.id for item in line_items)}",
                 True
             )
-            # raise err
-            return self.get_order_by_id(order_id)
+            raise err
 
     def retire_customer_anonymize_fields(
         self,
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -81,7 +81,7 @@ from commerce_coordinator.apps.commercetools.catalog_info.foundational_types imp
 from commerce_coordinator.apps.commercetools.utils import (
     find_latest_refund,
     find_refund_transaction,
-    get_refund_transaction_id_from_mobile_order,
+    get_refund_transaction_id_from_order,
     handle_commercetools_error,
     translate_refund_status_to_transaction_status
 )
@@ -559,10 +559,11 @@ class CommercetoolsAPIClient:
         try:
             logger.info(
                 f"[CommercetoolsAPIClient.update_return_payment_state_for_mobile_order] - "
-                f"Updating return payment state for return item(s) {return_line_item_return_ids} to 'REFUNDED'."
+                f"Updating return payment state for return item(s) "
+                f"{return_line_item_return_ids} to '{ReturnPaymentState.NOT_REFUNDED}'."
             )
 
-            refund_transaction_id = get_refund_transaction_id_from_mobile_order(order)
+            refund_transaction_id = get_refund_transaction_id_from_order(order)
             update_actions = []
 
             for return_item_id in return_line_item_return_ids:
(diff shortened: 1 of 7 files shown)


commerce_coordinator/apps/commercetools/clients.py
```diff
@@ -547,6 +548,60 @@ def update_return_payment_state_for_enrollment_code_purchase(
...
+            updated_order = self.base_client.orders.update_by_id(
+                id=order.id,
+                version=order.version,
+                actions=update_actions,
+            )
+
+            logger.info(
+                f"[CommercetoolsAPIClient.update_return_payment_state_for_mobile_order] - "
+                f"Successfully updated ReturnPaymentState to 'REFUNDED' for order ID: {order.id}. "
+                f"Updated return item IDs: {', '.join(return_line_item_return_ids)}."
```
syedsajjadkazmii, line 592: should add refund_transaction_id in log if available?

commerce_coordinator/apps/commercetools/utils.py
```diff
@@ -262,6 +262,26 @@ def find_latest_refund(payment: Payment):
...
+        str: The ID of the refund transaction, or an empty string if no refund transaction exists.
+    """
+    if order.payment_info and order.payment_info.payments:
+        for paymentReference in order.payment_info.payments:
+            payment = paymentReference.obj
+            for transaction in payment.transactions:
+                if transaction.type == TransactionType.REFUND:
+                    return transaction.id
+
+    return ''
```
NawfalAhmed, line 282: the util does nothing that is specific to mobile orders, so it can be a generic
`get_refund_transaction_id_from_order`

commerce_coordinator/apps/commercetools/utils.py
```diff
@@ -262,6 +262,26 @@ def find_latest_refund(payment: Payment):
...
+    Args:
+        order (Order): The Commercetools order object.
+
+    Returns:
+        str: The ID of the refund transaction, or an empty string if no refund transaction exists.
+    """
+    if order.payment_info and order.payment_info.payments:
+        for paymentReference in order.payment_info.payments:
+            payment = paymentReference.obj
+            for transaction in payment.transactions:
```
NawfalAhmed, line 278: we should only check in latest payment, and orders will always have payment_info so empty check is unnecessary
```suggestion
    for transaction in order.payment_info.payments[-1].obj:
```

NawfalAhmed (approved):

commerce_coordinator/apps/commercetools/utils.py
syedsajjadkazmii: the util flow is very specific to mobile orders, we are iterating through all the payments and all the transactions within each payment (ideally be one payment) just to check if there exist a refund transaction and return its id.

The flows  getting refund transactions for course purchase and program purchase would not be similar to this IMO. There can be failure payments, multiple refund payments etc for it. 

so I think this util is strictly written for mobile order  and should explicitly mention this in its name.

NawfalAhmed: > we are iterating through all the payments and all the transactions within each payment (ideally be one payment) just to check if there exist a refund transaction and return its id.
yea this is for any CT order, it doesn't have to be a mobile order.

only the purpose of using the util in ur specific code is tied to mobile order, not the util's logic, so IMO wouldn't name the util this way

NawfalAhmed: Up to you, in the end its just a name xD

NawfalAhmed: Here's an example, I recently I renamed an existing util function from `get_program_entitlements_standalone_prices` to
`get_standalone_prices_for_skus`
because I needed to use it in a different area unrelated to programs and the name suddenly didn't make sense
The util at time of creation may have been used in only for program entitlement sku but it's internal logic wasn't tied to programs, so it didn't need to be linked that way

commerce_coordinator/apps/commercetools/sub_messages/tasks.py
```diff
@@ -367,9 +367,13 @@ def _prepare_segment_event_properties(in_order, total_amount, return_line_item_r
     return_line_entitlement_ids = {return_line_items.get(line_item.id):
                                    get_line_item_lms_entitlement_id(line_item) for line_item in get_edx_items(order)}
 
+    is_mobile_order = False
+    if hasattr(order.custom, 'fields') and order.custom.fields:
```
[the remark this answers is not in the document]
syedsajjadkazmii: yes, thats better for sure. 

but just FYI, the commercetools's python SDK's Order Class always has `custom` field even though there doesn't exist in any field so the existing one is fine too. 

but I will go ahead and make the change

PR #447, opened 2025-06-20
refactor: added 30 minutes timeout of cache for order fulfillment webhook view

abdullahwaheed:


commerce_coordinator/apps/commercetools/views.py
```diff
@@ -57,7 +57,7 @@ def post(self, request):
 
         task_key = safe_key(key=order_id, key_prefix=tag, version='1')
 
-        if not acquire_task_lock(task_key):
+        if not acquire_task_lock(task_key, 1800):   # added 30 minutes cache timeout expiry for this
```
shafqatfarhan, line 60:
```suggestion
        if not acquire_task_lock(task_key, task_expiry=1800):   # added 30 minutes cache timeout expiry for this 
```

shafqatfarhan (approved):

commerce_coordinator/apps/commercetools/views.py
abdullahwaheed: we know the 2nd param is `task_expiry`, why should we explicitly need to name it?

PR #472, opened 2025-08-04
SONIC-1133: Fix refund amount in refund email for PSP

NawfalAhmed:
## Description

- fix refund amount in paypal and stripe refund events

## Supporting Information

Relevant JIRA Ticket: [SONIC-1133](https://2u-internal.atlassian.net/browse/SONIC-1133)

### DIFF
--- a/commerce_coordinator/apps/commercetools/sub_messages/tasks.py
+++ b/commerce_coordinator/apps/commercetools/sub_messages/tasks.py
@@ -386,7 +386,7 @@ def fulfill_order_returned_signal_task(order_id, return_items, message_id):
             else:
                 logger.info(f'[CT-{tag}] payment {psp_payment_id} refunded for message id: {message_id}')
 
-                total_in_dollars = result.get('amount_in_cents')
+                total_in_dollars = result.get('amount_in_dollars')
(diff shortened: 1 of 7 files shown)


commerce_coordinator/apps/commercetools/sub_messages/tasks.py
```diff
@@ -402,12 +386,15 @@ def _prepare_segment_event_properties(in_order, total_amount, return_line_item_r
             else:
                 logger.info(f'[CT-{tag}] payment {psp_payment_id} refunded for message id: {message_id}')
 
-                total_amount = result.get('amount_in_cents')
+                total_in_dollars = result.get('amount_in_cents')
```
syedsajjadkazmii, line 389: We should also rename `amount_in_cents` to `amount_in_dollars` or simply `refund_amount` because we are not working with cents from start (the function returning amount to refund returns `cents_to_dollars(amount)`) which means the amount was never in cents during pipeline execution.

I am not sure why it was named in cents but if we dont rename now, we will be confused again in future and would have to dry run to know again.

syedsajjadkazmii (approved):
not a blocking comment but would request you to handle before merging

PR #480, opened 2025-08-20
SONIC-1126: Show learners a message about the unavailability of 90-day-old orders

NawfalAhmed:
## Description

- restrict order history to 90 days

## Supporting Information

Relevant JIRA Ticket: [SONIC-1126](https://2u-internal.atlassian.net/browse/SONIC-1126)

### DIFF
--- a/commerce_coordinator/apps/frontend_app_ecommerce/views.py
+++ b/commerce_coordinator/apps/frontend_app_ecommerce/views.py
@@ -2,7 +2,8 @@
 Views for the frontend_app_ecommerce app
 """
 import logging
-from datetime import datetime
+from datetime import datetime, timedelta, timezone
+from itertools import chain, takewhile
 from typing import Union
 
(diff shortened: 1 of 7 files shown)


commerce_coordinator/apps/frontend_app_ecommerce/views.py
```diff
@@ -101,18 +107,30 @@ def get(self, request):
...
+                orders,
+                key=lambda order: date_conv(order["date_placed"]),
+                reverse=True,
+            )
 
-            output_orders = []
-
-            for order_set in order_data:
-                output_orders.extend(order_set['results'])
+            if params["cutoff_in_days"]:
```
shafqatfarhan, line 119: Why do we need this check since we are not passing this parameter in URL query params and instead setting it explicitly in a dict?

shafqatfarhan (approved):

PR #496, opened 2025-09-24
SONIC-1116: Allow staff to query order history for a specific learner

NawfalAhmed:
## Description

- allow staff to query order history for a specific learner

## Supporting Information

Relevant JIRA Ticket: [SONIC-1116](https://2u-internal.atlassian.net/browse/SONIC-1116)

### DIFF
--- a/commerce_coordinator/apps/commercetools/clients.py
+++ b/commerce_coordinator/apps/commercetools/clients.py
@@ -353,7 +353,6 @@ class CommercetoolsAPIClient:
             logger.info(f"[CommercetoolsAPIClient] - Customer found {user_message}")
             return results.results[0]
 
-    @conditional_retry
     def get_customer_by_lms_user_id(self, lms_user_id: int) -> Optional[Customer]:
         """
         Get a Commercetools Customer by their LMS User ID


commerce_coordinator/apps/frontend_app_ecommerce/views.py
```diff
@@ -81,17 +83,37 @@ def get(self, request):
...
+            params["username"] = learner_username
+            try:
+                learner = User.objects.get(username=learner_username)
+                params["email"] = learner.email
+                params["edx_lms_user_id"] = learner.lms_user_id
+            except User.DoesNotExist:
+                logger.info(
+                    "[UserOrdersView] User does not exist in Coordinator "
+                    "and hence doesn't have any Commercetools orders either"
+                )
```
shafqatfarhan, line 110: Are we continuing from here, whether the user is found or not?

commerce_coordinator/apps/frontend_app_ecommerce/views.py
```diff
@@ -81,17 +83,37 @@ def get(self, request):
...
+        if request.user.is_staff and learner_username:
+            params["username"] = learner_username
+            try:
+                learner = User.objects.get(username=learner_username)
+                params["email"] = learner.email
+                params["edx_lms_user_id"] = learner.lms_user_id
+            except User.DoesNotExist:
+                logger.info(
+                    "[UserOrdersView] User does not exist in Coordinator "
+                    "and hence doesn't have any Commercetools orders either"
```
shafqatfarhan, line 109: This is not true, actually, a CT user would always exist if that user has an order. There are a few scenarios now where the user does not exist in CC, but the order was created on CT. This was due to direct upselling URL to CT from the marketing site.

shafqatfarhan (approved):

commerce_coordinator/apps/frontend_app_ecommerce/views.py
```diff
@@ -101,8 +123,9 @@ def get(self, request):
             raise PermissionDenied(detail="Could not detect LMS user id.")
```
shafqatfarhan, line 123: Should not these checks also apply to the user in params now?

```suggestion
if not params.get("username"):  # pragma: no cover
# According to the Django checks this isnt possible with our current user model.
# Leaving in incase that changes.
raise PermissionDenied(detail="Could not detect username.")
if not params.get("edx_lms_user_id"):  # pragma: no cover
```

shafqatfarhan (approved):

commerce_coordinator/apps/frontend_app_ecommerce/views.py
NawfalAhmed: No,
This endpoints works for both support requests and normal order history requests,
If we don't receive username in request, we assume it's normal order history requests and returns the logged in user's orders.

shafqatfarhan: Exactly, that logged-in user would also be in params.get("username").

PR #557, opened 2026-08-12
feat: UPI payment_intent.succeeded webhook

bseverino:
## Summary

- [EDUN-15452](https://redventures.atlassian.net/browse/EDUN-15452): Make Stripe `payment_intent.succeeded` the authoritative async path for CommerceTools order finalization when UPI learners never return to `/payment-return`.
- Extends commerce-coordinator Stripe webhook ingress for `source_system=commercetools`, writing CT Payment/Order/fulfillment state and Segment `Order Completed` directly (parity with customer-twou finalize from [EDUN-15346](https://redventures.atlassian.net/browse/EDUN-15346)).
- Aligns with Schema B.4 ([Enable Payment Methods (UPI) - Schema & High-Level Technical Documentation](https://redventures.atlassian.net/wiki/spaces/EDU/pages/102974029879)): webhook finalizer lives in CC rather than Frontastic HTTP.
- Legacy `edx/commerce_coordinator?v=1` success path and `charge.refunded` handling remain unchanged.
- Scheduled orphan recovery is **out of scope**. Paid-without-enrollment when the webhook never completes is a support edge case (manual refund / replay); revisit if volume warrants automation.
- Known launch caveat until [EDUN-15347](https://redventures.atlassian.net/browse/EDUN-15347): dual-writer (browser finalize + webhook) may duplicate; accepted for this ship.

## What Was Built

| Component | Details |
| --- | --- |
| Stripe `WebhookView` | CT `payment_intent.succeeded` branch with signature verification + `SingleInvocation` on PI id; fires `payment_succeeded_commercetools_signal`. Event types compared to `StripeEventType.*.value`. If enqueue fails (`send_robust` receiver error), return **503** so Stripe retries and `handle_exception` clears the running flag (avoids ACK + 10-minute suppress). Refund SingleInvocation falls back to `event.id` when `idempotency_key` is null. |
| Shared finalize | Celery task → Charge txn (idempotent by `interactionId`) → Order from cart → pending fulfillment (from `Initial` by key) → Segment plan 18 (`is_mobile=False`, `payment_method=upi`) → PI metadata backfill (`order_id`, `ct_payment_id`) merged with existing keys. Existing-order path still heals `PENDING_FULFILMENT` + metadata. |
| CT client | `get_cart_by_id` (with `handle_commercetools_error`), `create_charge_payment_transaction`; `get_order_by_payment_id` raises `ValueError` when not found. Payment lookup by metadata id falls back to key only on `ResourceNotFound`. |
| Reliability | PI-scoped lock (5 min TTL, released in `finally`). Celery `max_retries=5` with **300s countdown** (same outage window as `fulfillment_completed_update_ct_line_item_task`); webhook ACKs Stripe only after successful enqueue, so Celery is the automated retry path for CT/Stripe errors. Structured quarantine logs for non-retryable failures. |

## Test Plan

- [x] Unit tests for CT routing (success / fail / legacy / unknown / SingleInvocation short-circuit / enqueue failure → 503 and running flag cleared)
- [x] Unit tests for shared finalize (happy path, Charge skip, already-existed fulfillment + metadata heal, CT outage propagate, `ResourceNotFound` vs transient, Segment props including `payment_method=upi`, lock contention)
- [x] Charge client unit test
- [ ] Stage smoke: succeed CT Stripe PI without `/payment-return` → CT Order + Charge + pending fulfillment + PI metadata + Segment
- [ ] Confirm legacy PI success + refund webhooks unchanged in stage

## Activation Checklist (ops)

- [ ] **Stripe Dashboard:** CC webhook endpoint subscribed to `payment_intent.succeeded` in stage then prod (owner / verified date). URL: `https://commerce-coordinator.<env>.edx.org/stripe/webhook/`

## NFR / Registration

- [x] N/A — existing edx commerce-coordinator Stripe webhook; no new EDU OpenAPI/AsyncAPI, EDU-Events, or QAaS registration in this change.

## Links

- Jira: [EDUN-15452](https://redventures.atlassian.net/browse/EDUN-15452) · Epic [EDUN-15262](https://redventures.atlassian.net/browse/EDUN-15262)
- Blocked by (parity): [EDUN-15346](https://redventures.atlassian.net/browse/EDUN-15346) · Blocks: [EDUN-15347](https://redventures.atlassian.net/browse/EDUN-15347)
- Schema: [Enable Payment Methods (UPI) - Schema & High-Level Technical Documentation](https://redventures.atlassian.net/wiki/spaces/EDU/pages/102974029879)


commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
```diff
@@ -0,0 +1,345 @@
...
+    order = client.create_order_from_cart(cart)
+
+    # --- Transition line items → PENDING_FULFILMENT ---
+    order = client.update_line_items_transition_state(
+        order_id=order.id,
+        order_version=order.version,
+        line_items=order.line_items,
+        from_state_id=order.line_items[0].state[0].state.id,
+        new_state_key=TwoUKeys.PENDING_FULFILMENT_STATE,
+        use_state_id=True,
```
rv-mromero: **Blocker:** this short-circuit heals PI metadata but skips line-item transition (and Segment).

If `create_order_from_cart` succeeded and `update_line_items_transition_state` then failed, Celery retry / recovery will see the order, return `already_existed=True`, and leave line items in `Initial`. That is paid + order exists + fulfillment never starts — not the EDUN-15347 duplicate-order case (CT still enforces one order per cart).

On this path, still transition line items that are not already `PENDING_FULFILMENT`, then backfill metadata. Segment can stay best-effort.

commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
```diff
@@ -0,0 +1,345 @@
...
+    )
+
+    # --- Emit Segment Order Completed (plan 18, web) ---
+    _emit_web_order_completed(client, order, cart, payment)
+
+    # --- Backfill PI metadata ---
+    _backfill_pi_metadata(
```
rv-mromero: `order.line_items[0].state[0].state.id` can `IndexError` if line items or state is empty, and it assumes every line item shares that `from` state.

Use `TwoUKeys.INITIAL_FULFILMENT_STATE` as `from_state` instead (built-in `Initial` key). That also makes the existing-order heal path safe to call when some items are already pending.

commerce_coordinator/apps/commercetools/tasks.py
```diff
@@ -552,3 +553,106 @@ def revoke_line_mobile_order_task(payment_id: str):
...
+            "source": source,
+        },
+    )
+
+
+@shared_task(
+    bind=True,
+    autoretry_for=(CommercetoolsError, stripe.error.StripeError),
+    max_retries=5,
+    retry_kwargs={"countdown": 3},
```
rv-mromero: `countdown=3` × `max_retries=5` is ~15s. `fulfillment_completed_update_ct_line_item_task` in this file uses `countdown=300` specifically to cover CT outages.

Webhook `SingleInvocation` TTL is 10 minutes (`NOTIFICATION_CACHE_TTL_SECS`), so Stripe's ~5-minute retry is dropped and the next delivery is typically ~1 hour. Recovery cron is the real backstop — please either lengthen this countdown or call out in the PR that merge ≠ launch until cron is scheduled.

commerce_coordinator/apps/commercetools/management/commands/recover_orphaned_stripe_commercetools_payments.py
```diff
@@ -0,0 +1,341 @@
...
+        next_page = None
+
+        while has_more and len(orphan_ids) < limit:
+            kwargs = {"query": query, "limit": min(100, limit - len(orphan_ids))}
+            if next_page:
+                kwargs["page"] = next_page
+
+            result = stripe.PaymentIntent.search(**kwargs)
+
+            for pi in result.data:
```
rv-mromero: Stripe Search docs use double quotes: `status:"succeeded" AND metadata["source_system"]:"commercetools"`.

Single quotes often work, but if Search rejects this query the `except` in `_discover_stripe_orphans` falls through to `_list_filter_stripe_orphans`, which pages **every** PI since `--since` until it finds `--limit` orphans. Please confirm the query in stage, and cap pages/PIs examined on the list fallback.

commerce_coordinator/apps/commercetools/management/commands/recover_orphaned_stripe_commercetools_payments.py
```diff
@@ -0,0 +1,341 @@
...
+            except (CommercetoolsError, Exception) as exc:
+                self.stderr.write(f"  [quarantine] {pi_id}: {exc}")
+                _log_quarantine(
+                    pi_id=pi_id,
+                    ct_payment_id=meta.get("ct_payment_id") or "unknown",
+                    ct_cart_id=meta.get("ct_cart_id") or "unknown",
+                    reason=str(exc),
+                    source="recovery",
+                )
+                quarantined += 1
```
rv-mromero: Webhook Celery retries `CommercetoolsError` / `StripeError`; this command quarantines on the first transient.

Next cron will retry, but a New Relic alert on `[quarantine]` will page during blips. Either retry a few times in-command, or don't call `_log_quarantine` for retryable errors (log a warning instead).

rv-mromero (changes requested):
## Request changes

Architecture matches [EDUN-15452](https://redventures.atlassian.net/browse/EDUN-15452): Stripe-signed `payment_intent.succeeded` for `source_system=commercetools`, shared CT finalize, recovery command, legacy/refund paths left intact. Copilot's earlier comments (metadata merge, `ResourceNotFound`-only fallback, `max_retries` on the decorator, refund `event.id`) are addressed. CI is green.

The gap is **partial success**: if the order is created and a later step fails, retry/recovery will not finish fulfillment.

### Blocker: existing-order path skips fulfillment

`finalize_ct_order_from_stripe_pi` creates the order, then transitions line items to `PENDING_FULFILMENT`, then emits Segment, then backfills PI metadata. If anything after `create_order_from_cart` fails (CT blip, empty `line_items`/`state`, worker kill), the next run hits `get_order_by_payment_id` and returns `already_existed=True` without retrying the transition.

Same-cart double-create is naturally blocked by CommerceTools (one order per cart), so this is not the EDUN-15347 duplicate-order case. It is **paid + order exists + line items still Initial**, so enrollment may never start. Recovery will keep treating it as "already existed" and skip.

**Fix:** on the existing-order path, still transition line items that are not already pending (use `TwoUKeys.INITIAL_FULFILMENT_STATE` as `from_state`, not `line_items[0].state[0]`). Keep Segment best-effort if you want parity, but the state heal is the one that unblocks learners.

### Medium

- **Celery retries vs SingleInvocation TTL.** New task: `max_retries=5`, `countdown=3` (~15s). Fulfillment tasks in this file wait **5 minutes** between retries. Webhook `SingleInvocation` cache TTL is **10 minutes**, so Stripe's 5-minute retry is dropped; the next real webhook retry is typically ~1 hour. Recovery is the real backstop — do not treat merge as launch-ready until the cron in the PR checklist is actually scheduled.
- **Recovery Search fallback can scan the whole Stripe account.** Confirm the Search query in stage (docs show double quotes). If Search rejects the query, `_list_filter_stripe_orphans` pages every PI since `--since` until it finds `--limit` orphans. Cap pages/PIs examined.
- **Recovery quarantines transients immediately.** Webhook Celery retries CT/Stripe errors; the management command logs `[quarantine]` on the first `CommercetoolsError`. Next cron will retry, but a noisy New Relic alert on that contract will page during blips. Either retry a few times in-command, or don't quarantine retryable errors.
- **No lock around finalize.** Webhook + recovery can both enter create-order. CT will reject the second cart→order, which is fine, but concurrent Charge adds can still duplicate `interactionId`s. A short `acquire_task_lock` on PI id (same pattern as `fulfillment_completed_update_ct_line_item_task`) would serialize this ticket's own writers. EDUN-15347 still owns browser+webhook analytics duplication.

### Low / nits

- `get_cart_by_id` has no `handle_commercetools_error` wrapper (unlike `create_order_from_cart`).
- Extra Stripe `PaymentIntent.retrieve` per orphan in recovery before finalize (Copilot); fetch metadata only on quarantine.
- Coverage: new finalize ~74%, recovery command ~76%; missing task retry-exhaustion and empty line-item paths.
- Drive-by: deleted the `SONIC-898` IP-allowlist TODO on `WebhookView`. Unrelated; keep it or replace with a ticket.

### What looks solid

- Signature check on raw body; CT vs legacy `source_system` split; unknown sources no-op 200.
- Charge txn skipped when `interaction_id` already matches; CT outage on order lookup is not treated as not-found.
- PI metadata backfill merges existing keys (`source_system`, `ct_cart_id`).
- Refund SingleInvocation falls back to `event.id` when `idempotency_key` is null.
- Segment payload: plan 18, `is_mobile=False`, `payment_method` from CT (`upi` in tests), totals from USD cart.
- Tests cover routing, Charge skip, metadata heal, `ResourceNotFound` vs transient, recovery dry-run / fallback / CT-secondary.

**Launch still depends on ops**, even after the code fix: Stripe Dashboard subscribed to `payment_intent.succeeded`, recovery cron, quarantine alert. Stage smokes in the PR body are still unchecked.

rv-mromero (approved):

rv-mromero (approved):

commerce_coordinator/apps/commercetools/management/commands/recover_orphaned_stripe_commercetools_payments.py
```diff
@@ -0,0 +1,398 @@
+"""
+Management command to discover and finalize orphaned Stripe PaymentIntents /
+CommerceTools Payments that have succeeded without a linked Order.
```
colinbrash: @bseverino am I right that the orphan case should be extremely rare, basically only because of a CT outage?  The webhook (`payment_intent.succeeded`) should handle the cases where it is just the user not returning?

If so, I don't think we need this command at all, we can just handle these are manual-refund support cases should they ever happen.

commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
```diff
@@ -0,0 +1,421 @@
...
+from commerce_coordinator.apps.commercetools.clients import CommercetoolsAPIClient
+from commerce_coordinator.apps.core.memcache import safe_key
+from commerce_coordinator.apps.core.segment import track
+from commerce_coordinator.apps.core.tasks import acquire_task_lock, release_task_lock
+
+logger = logging.getLogger(__name__)
+
+FINALIZE_LOCK_PREFIX = "finalize_ct_order_from_stripe_pi"
+# Stripe + multiple CT calls + Segment can exceed the default 60s lock TTL.
+FINALIZE_LOCK_EXPIRE = 1800  # 30 minutes; matches commercetools/views.py
```
colinbrash: this feels like a long time for a crashed worked to block other workers.  is 30m really a standard, or are we continuing something excessive?

commerce_coordinator/apps/commercetools/management/commands/recover_orphaned_stripe_commercetools_payments.py
bseverino: Agreed, dropping the command and we can treat as an edge case

commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
bseverino: Default lock TTL is 60s, 1800s was copied from `fulfill-order-placed`. 60s might still be too short, so lowered it to 5 min

colinbrash (approved):

colinbrash:
without the recovery command is the celery timeout the right length?  not sure if we wanted to adjust it to be closer to the fulfillment CT update task (longer in case of a CT outage).  approving either way 👍

PR #559, opened 2026-08-20
fix: recover overlapping Stripe webhook order creation

bseverino:
## Summary

- When the Stripe webhook finalizer fails to create a CommerceTools order because the browser path already converted the cart, look the order up again and finish metadata/fulfillment instead of retrying create or treating it as a hard failure ([EDUN-15347](https://redventures.atlassian.net/browse/EDUN-15347)).
- Do not emit a second `Order Completed` event when that existing order is found. If the lookup itself fails in CommerceTools, let Celery retry that error instead of hiding it behind the original create error.
- Implements the commerce-coordinator half of production control 5 from [Enable Payment Methods (UPI) - Schema & High-Level Technical Documentation](https://redventures.atlassian.net/wiki/spaces/EDU/pages/102974029879) (PRD appendix B.5). Pair with the customer-twou PR before UPI production traffic.

## What Was Built

| Component | Details |
| --- | --- |
| `finalize_ct_order_from_stripe_pi` | After `create_order_from_cart` raises `CommercetoolsError`, query `get_order_by_payment_id` again |
| Existing-order heal | If an order is found, run the existing heal path (PENDING_FULFILMENT + PaymentIntent metadata) and return `already_existed=True` with no Segment event |
| Empty lookup | `ValueError` (no order) re-raises the original create error so bounded retries / recovery logging still apply |
| Lookup outage | Re-query `CommercetoolsError` propagates so Celery retries the lookup failure |

## Test Plan

- [x] Create fails and an order exists → heal, no `track`, `already_existed=True`
- [x] Create fails and lookup returns not-found → original create error is re-raised
- [x] Create fails and lookup raises `CommercetoolsError` → lookup error propagates
- [x] Existing charge-skip / find-before-create / recovery-log coverage still passes
- [ ] Stage: overlapping `/payment-return` + `payment_intent.succeeded` → one CT order, one Charge txn, one Segment event

## NFR Compliance

- Not an EDU NestJS public API; existing Stripe webhook → Celery finalize path only.
- No new endpoints, events, or service-discovery registration.
- Recovery logging contract is unchanged (`quarantine`, `pi_id`, `ct_payment_id`, `ct_cart_id`, `reason`, `source`).

### DIFF
--- a/commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
+++ b/commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
@@ -329,7 +329,7 @@ def _finalize_ct_order_from_stripe_pi_locked(
     except CommercetoolsError as create_error:
         try:
             existing_order = client.get_order_by_payment_id(payment.id)
-        except (ValueError, CommercetoolsError):
+        except ValueError:
             existing_order = None
         if existing_order is None:
             raise create_error
--- a/commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
+++ b/commerce_coordinator/apps/commercetools/stripe_payment_finalize.py
@@ -160,6 +160,24 @@ def _ensure_pending_fulfilment(client, order):
     )
 
 
+def _heal_existing_order(client, order, payment_intent_id, payment, metadata):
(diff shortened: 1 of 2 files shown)


colinbrash (approved):
:shipit:

